PR #8, opened 2020-09-24
Log the exception when failing to map nms argument types, don't register UUID argument before 1.16

jmanpenilla:
the nms UUID argument type doesn't exist until 1.16.


Sauilitired:
Thanks! I didn't even realise they didn't include of the argument types initially. This is a way cleaner solution than letting the entire thing break.

PR #12, opened 2020-09-28
Switch to gradle

Sauilitired:
Also switch from jsr305 to checker-qual annotations and from Guava to geantyref.


build.gradle
```diff
@@ -0,0 +1,130 @@
+buildscript {
+    repositories {
+        mavenCentral()
+        maven { url = "https://oss.sonatype.org/content/repositories/snapshots/" }
+        jcenter()
+    }
+    dependencies {
+        classpath("com.github.jengelman.gradle.plugins:shadow:5.0.0")
+        classpath 'com.bmuschko:gradle-nexus-plugin:2.3.1'
```
N0tMyFaultOG, line 8:
```suggestion
        classpath("com.bmuschko:gradle-nexus-plugin:2.3.1")
```

broccolai: Single quotes are used in most places, I think double quotes are better looking though so might wanna change the rest too

PR #56, opened 2020-10-16
Fix breaking changes in Velocity

STG-Allen:
CommandSource#sendMessage now requires an Identity


jmanpenilla (approved):

Sauilitired:
Thanks for the fix!

PR #59, opened 2020-10-16
Add an adventure text colour parser

Sauilitired:
**Untested** Should probably be tested before merging.

PR #60, opened 2020-10-16
Captions for Entity Selecors and repeated "No input was provided'

jmanpenilla:


broccolai (dismissed):
Haven't tested but it looks good, we should add the colours argument from #59 too

PR #63, opened 2020-10-17
:sparkles: Add LocationArgument to cloud-bukkit

Sauilitired:


Sauilitired:
Suggestions are broken without Brigadier at the moment, so that needs toe be fixed before merging.

cloud-minecraft/cloud-bukkit/src/main/java/cloud/commandframework/bukkit/parsers/location/LocationArgument.java
```diff
@@ -0,0 +1,317 @@
...
+                );
+                if (coordinate.getFailure().isPresent()) {
+                    return ArgumentParseResult.failure(
+                            coordinate.getFailure().get()
+                    );
+                }
+                coordinates[i] = coordinate.getParsedValue().orElseThrow(NullPointerException::new);
+            }
+            final Location originalLocation;
+            final CommandSender mappedSender = this.bukkitCommandManager.getBackwardsCommandSenderMapper()
```
jmanpenilla, line 208: `final CommandSender mappedSender = commandContext.get("BukkitCommandSender");` and then we don't need to supply a command manager to the builder. This is added to the contexts by the BukkitCommandPreprocessor.

jmanpenilla (changes requested):

PR #73, opened 2020-10-18
Fix multiple chained optionals not working correctly

Sauilitired:
There is a problem where the child arguments never forwarded their commands correctly. This will now fix itself when necessary.


Sauilitired:
This needs to be added to the changelog as well. Whoops.

JarFiles (approved):

PR #83, opened 2020-10-22
Make the flag parser smarter

### LINKED ISSUE #75: Better flags
- Allow joining presence alias, like `-a -b -c` <=> `-abc`
- Look into feasibility of creating literal chains for flags (when under some threshold, as this adds up to n! + (n - 1)! + (n - 2) + ... + 1! literals when using Brigadier <-- Check how Sponge does it

Forcing flags to the end of the command is a little annoying, but because of the way Brigadier works, we're quite limited in how we do this, if we want the same exact syntax for brig vs. no brig.

Sauilitired:
It will now allow multiple presence flag aliases to be joined into a single flag, such that `-a -b -c <=> -abc`.

This fixes #75

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,7 +21,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
  - Added CaptionKeys to cloud-bungee
  - Added BungeeCommandPreprocessor to cloud-bungee
 
-### Chanced
+### Changed
  - Allow for combined presence flags, such that `-a -b -c` is equivalent to `-abc`
 
 ## [1.0.2] - 2020-10-18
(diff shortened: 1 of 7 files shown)


Sauilitired:
This needs more testing (preferably in-game!)

CHANGELOG.md
```diff
@@ -21,6 +21,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
  - Added CaptionKeys to cloud-bungee
  - Added BungeeCommandPreprocessor to cloud-bungee
 
+### Chanced
```
jmanpenilla, line 24: Changed

jmanpenilla (approved):
tested briefly in game with a couple presence flags with/without brig enabled

PR #97, opened 2020-10-24
:sparkles: Allow for recursive annotations

Sauilitired:
This allows people to make annotated annotations, that contain default perms, etc.

### DIFF
--- a/cloud-annotations/src/main/java/cloud/commandframework/annotations/AnnotationParser.java
+++ b/cloud-annotations/src/main/java/cloud/commandframework/annotations/AnnotationParser.java
@@ -116,10 +116,8 @@ public final class AnnotationParser<C> {
     ) {
         A innerCandidate = null;
         for (final Annotation annotation : annotations) {
-            if (checkedAnnotations.contains(annotation.annotationType())) {
+            if (!checkedAnnotations.add(annotation.annotationType())) {
                 continue;
-            } else {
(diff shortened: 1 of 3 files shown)


Proximyst (changes requested):
```java
@Anno2
@interface Anno1 {}

@Anno1
@interface Anno2 {}
```

Oops! This should be protected against, as some annotations (e.g. checkerframework? javax?) might do this.

Sauilitired:
Hm, that is fair. The method could take in a set of already checked annotation classes and ignore those

broccolai (approved):

cloud-annotations/src/main/java/cloud/commandframework/annotations/AnnotationParser.java
```diff
@@ -106,16 +108,42 @@ public AnnotationParser(
...
+            final @NonNull Class<A> clazz,
+            final @NonNull Set<Class<? extends Annotation>> checkedAnnotations
+    ) {
+        A innerCandidate = null;
+        for (final Annotation annotation : annotations) {
+            if (checkedAnnotations.contains(annotation.annotationType())) {
+                continue;
+            } else {
+                checkedAnnotations.add(annotation.annotationType());
+            }
```
Proximyst, line 123:
```suggestion
            if (!checkedAnnotations.add(annotation.annotationType())) {
                continue;
            }
```

PR #109, opened 2020-10-26
✨ Throw AmbiguousNodeException for more ambiguous cases

jmanpenilla:
AmbiguousNodeException will now be thrown when ambiguous commands such as
``/command <player>`` and ``/command <location>`` are attempted to be registered.


Sauilitired:
do we wanna add this to the changelog or nah? other than that it do be looking good

jmanpenilla:
> do we wanna add this to the changelog or nah? other than that it do be looking good

sure why not

PR #114, opened 2020-10-27
🐛 Fix Bukkit alias command suggestions without Brigadier

jmanpenilla:


broccolai (approved):
Looks good in my tests

PR #122, opened 2020-10-29
:bug: Fix number suggestions when there are following arguments

jmanpenilla:
Adds a test to ensure `command <number>` 
and `command <number> <another_argument>`
have the same exact suggestions for the `<number>` argument.

### DIFF
--- a/cloud-core/src/test/java/cloud/commandframework/CommandSuggestionsTest.java
+++ b/cloud-core/src/test/java/cloud/commandframework/CommandSuggestionsTest.java
@@ -23,6 +23,7 @@
 //
 package cloud.commandframework;
 
+import cloud.commandframework.arguments.standard.BooleanArgument;
 import cloud.commandframework.arguments.standard.EnumArgument;
 import cloud.commandframework.arguments.standard.IntegerArgument;
 import cloud.commandframework.arguments.standard.StringArgument;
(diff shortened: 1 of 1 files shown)


cloud-core/src/test/java/cloud/commandframework/CommandSuggestionsTest.java
```diff
@@ -206,6 +210,23 @@ void testNumbers() {
...
+        Assertions.assertEquals(Arrays.asList("0", "1", "2", "3", "4", "5", "6", "7", "8", "9"), suggestions);
+        final String input2 = "numberswithfollowingargument 1";
+        final List<String> suggestions2 = manager.suggest(new TestCommandSender(), input2);
+        Assertions.assertEquals(Arrays.asList("1", "10", "11", "12", "13", "14", "15", "16", "17", "18", "19"), suggestions2);
+        final String input3 = "numberswithfollowingargument -";
+        final List<String> suggestions3 = manager.suggest(new TestCommandSender(), input3);
+        Assertions.assertEquals(Arrays.asList("-1", "-2", "-3", "-4", "-5", "-6", "-7", "-8", "-9"), suggestions3);
+        final String input4 = "numberswithfollowingargument -1";
+        final List<String> suggestions4 = manager.suggest(new TestCommandSender(), input4);
+        Assertions.assertEquals(Arrays.asList("-1", "-10", "-11", "-12", "-13", "-14", "-15", "-16", "-17", "-18", "-19"),
```
bergerkiller, line 226: I don't believe '-1' should be included as suggestion, since this was already typed by the user. In all other cases (such as literals) the exact matching suggestions aren't shown.

jmanpenilla: No, this is intended. The tests in `testNumbersWithFollowingArguments` are almost exactly the same as the tests in `testNumbers` (which pass), with the only difference being that there is an additional argument following the integer argument rather than the integer argument being the last one.
`command <number>` vs `command <number> <another_argument>`

PR #126, opened 2020-10-30
Improve extensibility of the MinecraftExceptionHandler

jmanpenilla:
This lets you have access to the sender without casting the exception and getting it from the command context


Proximyst (approved):
LGTM :rocket:

Issue #137, opened 2020-11-16
Pattern matching where literals are mixed with arguments


bergerkiller:
I don't know if this is a bug, something that is purposefully not implemented, or something I could work around using annotations I've missed in the documentation.

I have two commands declared:
```
    @CommandMethod("command literal")
    @CommandDescription("Performs the literal operation")
    private void commandLiteral(
              final CommandSender sender
    ) {
        sender.sendMessage("Performed command with a literal match");
    }

    @CommandMethod("command <argument>")
    @CommandDescription("Performs the operation with some variable")
    private void commandArgument(
              final CommandSender sender,
              final @Argument(value="argument") String argument
    ) {
        sender.sendMessage("Performed command using " + argument);
    }
```

But get the error:
```
com.bergerkiller.bukkit.tc.dep.cloud.exceptions.AmbiguousNodeException: Ambiguous Node: argument cannot be added as a child to command (All children: literal, argument)
    at com.bergerkiller.bukkit.tc.dep.cloud.CommandTree.checkAmbiguity(CommandTree.java:741)
    at java.lang.Iterable.forEach(Iterable.java:75)
    at com.bergerkiller.bukkit.tc.dep.cloud.CommandTree.checkAmbiguity(CommandTree.java:745)
    at com.bergerkiller.bukkit.tc.dep.cloud.CommandTree.verifyAndRegister(CommandTree.java:672)
    at com.bergerkiller.bukkit.tc.dep.cloud.CommandTree.insertCommand(CommandTree.java:617)
    at com.bergerkiller.bukkit.tc.dep.cloud.CommandManager.command(CommandManager.java:210)
    at com.bergerkiller.bukkit.tc.dep.cloud.annotations.AnnotationParser.parse(AnnotationParser.java:250)
```

I want /command literal to match one, and have /command anythingelse match the other. Obviously I can handle this by doing an equals check of the argument, but let's assume there are a lot of literals like this. Or that some literals have flags, while the variable one doesn't.

Is there a solution?

jmanpenilla:
An `AmbiguousNodeException` is thrown here because by cloud's definition, those nodes are ambiguous. A node can have multiple literals, or a single argument as children. See #130 and #106 for further context.

bergerkiller:
Got it

PR #142, opened 2020-11-21
Add support for literal arguments overriding dynamic ones

bergerkiller:
This is a proposed fix for issue https://github.com/Incendo/cloud/issues/137

Prerequisite pulls:
- https://github.com/Incendo/cloud/pull/146
- https://github.com/Incendo/cloud/pull/147

**TODO: Brigadier needs to be updated to supply the right suggestions**

I've added support for literals overriding variable arguments in the command tree. In the past this threw ambiguity exceptions, I've rewritten the ambiguity detection to be smarter. It was checking whether the number of children was > 1 if any of the arguments was not static, which broke a lot of things.

Now it allows only one variable (parsed) argument as a child, but allows multiple literal static arguments to co-exist. When parsing the command, it will first check for the literal arguments (and aliases). If none of those match, then it will parse variable ones.

This adds support for command registration such as:
```
/command literal help
/command <variable> help
```

Which when invoked as:
`/command literal help`

Will execute the first, while invoked as:
`/command arg help`

Executes the other. Added a test for this as well.

In addition, I've had to fix a problem in the code itself that occurs when running tests. When doing the ambiguity test, it checks if an exception is thrown when ambiguity occurs. However, the tree does not clean up the failing command, so it is left in an invalid state. This caused all later-running tests to fail as well. As a result it heavily depended on test execution order whether all the tests succeeded. I fixed this by cleaning up some of the added node children upon failure.

### DIFF
--- a/cloud-core/src/main/java/cloud/commandframework/CommandTree.java
+++ b/cloud-core/src/main/java/cloud/commandframework/CommandTree.java
@@ -470,20 +470,24 @@ public final class CommandTree<C> {
             final @NonNull Queue<@NonNull String> commandQueue,
             final @NonNull Node<@Nullable CommandArgument<C, ?>> root
     ) {
-
         /* If the sender isn't allowed to access the root node, no suggestions are needed */
         if (this.isPermitted(commandContext.getSender(), root) != null) {
             return Collections.emptyList();
         }
         final List<Node<CommandArgument<C, ?>>> children = root.getChildren();
-        if (children.size() == 1 && !(children.get(0).getValue() instanceof StaticArgument)) {
-            return this.suggestionsForDynamicArgument(commandContext, commandQueue, children.get(0));
-        }
-        /* There are 0 or more static arguments as children. No variable child arguments are present */
-        if (children.isEmpty() || commandQueue.isEmpty()) {
-            return Collections.emptyList();
-        } else {
-            final Iterator<Node<CommandArgument<C, ?>>> childIterator = root.getChildren().iterator();
+
+        /* Calculate a list of arguments that are static literals */
+        final List<Node<CommandArgument<C, ?>>> staticArguments = children.stream()
+                .filter(n -> n.getValue() instanceof StaticArgument)
+                .collect(Collectors.toList());
+
+        /*
+         * Try to see if any of the static literals can be parsed (matches exactly)
+         * If so, enter that node of the command tree for deeper suggestions
+         */
+        if (!staticArguments.isEmpty() && !commandQueue.isEmpty()) {
+            final Queue<String> commandQueueCopy = new LinkedList<String>(commandQueue);
+            final Iterator<Node<CommandArgument<C, ?>>> childIterator = staticArguments.iterator();
             if (childIterator.hasNext()) {
                 while (childIterator.hasNext()) {
                     final Node<CommandArgument<C, ?>> child = childIterator.next();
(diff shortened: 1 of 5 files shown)


bergerkiller:
One problem I've found is that the suggestions will only show the literals, not the variable-argument commands that also exist. If that is something I should fix first, let me know.

jmanpenilla:
This type of behavior is not intended to be supported, and so it's not likely this will be merged. For it to be considered however, it would definitely need to have working suggestions, and it would need to take into account Brigadier.

bergerkiller:
That is fair, I'll keep it in-house then since it is very useful in my personal situation. When I fix the suggestion problems, I'll amend this PR with the relevant commits, unless this PR is closed by then.

Proximyst:
Pardon my inability to test on my own: given a tree such that
```
/help
      | <arg> help
      | literal help
```
show the 2nd option as the 2nd option when typed into the client? If it does not, I don't believe we should consider this given Brigadier itself does not (though we would be open to it if a PR was opened to Brigadier to fix such an issue!). If it *does* however choose to show the 2nd option to the client where it's supposed to (and the 1st where it's supposed to), I'd love to :+1: this PR.

bergerkiller:
When you type:
`/help lit`
It will suggest 'literal' but none of the suggestions that <arg> would give (if a suggestion provider is used)

When you type:
`/help a`
It will suggest the suggestions of <arg> (if any start with 'a')

When you type:
`/help l`
It will suggest only 'literal', not any of the suggestions provided by <arg> that start with l.

Correct me if I'm wrong, but I believe desired is that:
`/help `
Will show '<arg>' and all the literals in a single list

`/help l`
Will concatenate the 'literal' suggestion with any other suggestion provided by the <arg> argument

bergerkiller:
Yeah this needs more work, also for how I'm using it myself, I'll post an update when things are fixed

bergerkiller:
I've fixed a lot of issues that happened because of https://github.com/Incendo/cloud/issues/144

Is this how it should work? I'll work on brigadier support next.

<details>
    <summary>Example code</summary>

```java
        manager.command(
                manager.commandBuilder("command", Description.of("Test cloud command using a builder"))
                        .argument(StringArgument.<CommandSender>newBuilder("argument").withSuggestionsProvider((context, inputQueue) -> {
                            return Arrays.asList("one", "two", "link");
                        }).build())
                        .literal("test")
                        .handler(context -> {
                            context.getSender().sendMessage("Argument: " + context.get("argument"));
                        })
        );

        manager.command(
                manager.commandBuilder("command", Description.of("Test cloud command using a builder"))
                        .literal("literal")
                        .handler(context -> {
                            context.getSender().sendMessage("Literal command executed");
                        })
        );
```
</details>

Video of it in action:
https://www.dropbox.com/s/6fpg2y7vhs0ar4w/2020-11-22%2021-12-03.mp4?dl=0

It combines the suggestions from the literals with that of the variable argument, automatically hiding suggestions that don't apply (don't start with what the user typed)

Sauilitired:
A major problem with this is that there's no indications in existing contracts of the parsing order of nodes in the command tree. This behaviour is dependent on whether or not literal arguments are parsed before variable arguments, and existing contracts would need to be strengthened. Whether or not that should be done is something that needs to be discussed, as it will impact how the command tree code evolves.

Concatenating suggestions is quite awkward, but very necessary if this is the route we want to take. That shouldn't really have any significant impact on current suggestion providers, despite them being written under the assumption that they would be the sole suggestion providers for any given argument.

Do these changes allow for anything that wouldn't be possible with a custom argument type/string argument with custom suggestions?

bergerkiller:
`Do these changes allow for anything that wouldn't be possible with a custom argument type/string argument with custom suggestions?`

Yes, multiple variable tree nodes assigned to different literals. For example:
```
/command literal subcommanda
/command literal subcommandb
/command othercommand --flag hello
/command <variable> help
```

Can validly be registered with my changes, but break before. You would need to parse all variables (greedy), which breaks suggestions and renders the entire cloud api useless for everything under /command.

What I did prior to these fixes is store the full argument list using a preprocessor, and simply used an @Argument("name") @Greedy to parse anything. This was a placeholder, because this broke all suggestions.

bergerkiller:
I've fixed the duplicate suggestion problem with brigadier. One thing I don't really understand is why the brigadier argument names don't show anymore. The same happens when testing a command that doesn't mix arguments with literals. Seems almost like this thing broke entirely, because it does register the <argument> or [flags] literal in the brigadier manager.

Could use some advice, I don't know that much about brigadier :<

Sauilitired:
This was not meant to be closed, I'm sorry D: I thought it'd let me change the target branch after 1.2.0-dev had been merged, but it doesn't... Thanks GitHub.

PR #143, opened 2020-11-22
:bug: Annotations API fix

bergerkiller:
When registering a command such as this:
```java
    @CommandMethod("train ticket <ticket>")
    private void commandGiveTicket(
              final CommandSender sender,
              final @Argument("ticket") String ticketName
    ) {
    }
```

This caused a weird situation where the ticketName argument was bound to the first ticket literal, rather than the second. This broke some things when you had other literals defined on the same level. I tracked it down to how the syntax fragments are parsed, which used a LinkedHashMap. This is not correct, since literals can repeat as argument names. Quite common to do this.

This commit replaces that with a List<SyntaxFragment> and a simple lookup function which only looks for arguments, skipping literals entirely. Literals should never match with @Argument parameters of the method.

### DIFF
--- a/cloud-annotations/src/main/java/cloud/commandframework/annotations/AnnotationParser.java
+++ b/cloud-annotations/src/main/java/cloud/commandframework/annotations/AnnotationParser.java
@@ -55,7 +55,6 @@ import java.util.Arrays;
 import java.util.Collection;
 import java.util.HashMap;
 import java.util.HashSet;
-import java.util.LinkedHashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.Optional;
@@ -261,7 +260,7 @@ public final class AnnotationParser<C> {
         for (final CommandMethodPair commandMethodPair : methodPairs) {
             final CommandMethod commandMethod = commandMethodPair.getCommandMethod();
             final Method method = commandMethodPair.getMethod();
-            final LinkedHashMap<String, SyntaxFragment> tokens = this.syntaxParser.apply(commandMethod.value());
+            final List<SyntaxFragment> tokens = this.syntaxParser.apply(commandMethod.value());
             /* Determine command name */
             final String commandToken = commandMethod.value().split(" ")[0].split("\\|")[0];
             @SuppressWarnings("rawtypes") final CommandManager manager = this.manager;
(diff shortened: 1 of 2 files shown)


Sauilitired (approved):

cloud-annotations/src/main/java/cloud/commandframework/annotations/AnnotationParser.java
```diff
@@ -285,27 +284,27 @@ public AnnotationParser(
             for (final ArgumentParameterPair argumentPair : arguments) {
                 final CommandArgument<C, ?> argument = this.buildArgument(
                         method,
-                        tokens.get(argumentPair.getArgument().value()),
+                        findSyntaxFragment(tokens, argumentPair.getArgument().value()),
```
jmanpenilla, line 287:
```suggestion
                        this.findSyntaxFragment(tokens, argumentPair.getArgument().value()),
```

cloud-annotations/src/main/java/cloud/commandframework/annotations/AnnotationParser.java
```diff
@@ -372,6 +371,18 @@ public AnnotationParser(
         return commands;
     }
 
+    private @NonNull SyntaxFragment findSyntaxFragment(final @NonNull List<SyntaxFragment> fragments,
```
jmanpenilla, line 374:
```diff
-     private @NonNull SyntaxFragment findSyntaxFragment(final @NonNull List<SyntaxFragment> fragments,
-                                                        final @NonNull String argumentName
+     private @NonNull SyntaxFragment findSyntaxFragment(
+             final @NonNull List<SyntaxFragment> fragments,
+             final @NonNull String argumentName
+     ) {
```

jmanpenilla (changes requested):
LGTM, just some minor codestyle things

bergerkiller:
I'm a little new to github's auto-resolving feature. If I use the 'resolve conversation' button, does that add new commits to my repository with your suggested changes, only to this PR, or something else altogether? Ideally I keep all changes inside a single commit, but I could squash it later.

Proximyst:
Resolving just collapses the comment for everyone.

bergerkiller:
Ah, so commit suggestion button makes the new commit most likely. Does that commit to my own (bergerhealer) repo?

Sauilitired:
> Ah, so commit suggestion button makes the new commit most likely. Does that commit to my own (bergerhealer) repo?

it makes a commit to the source branch, which in this case is in your repository

bergerkiller:
Applied the requested changes

jmanpenilla (approved):

Issue #144, opened 2020-11-22
(String) argument suggestions do not show up because argument succeeds parsing


bergerkiller:
I didn't know whether to put this in the PR or as a separate issue, so I'm putting it out here.

**Note: no brigadier is used**

https://github.com/Incendo/cloud/blob/1.2.0-dev/cloud-core/src/main/java/cloud/commandframework/CommandTree.java#L508

In the CommandTree suggestions code, the following is done when providing suggestions for non-static arguments:
```java
            // START: Preprocessing
            final ArgumentParseResult<Boolean> preParseResult = child.getValue().preprocess(
                    commandContext,
                    commandQueue
            );
            if (preParseResult.getFailure().isPresent() || !preParseResult.getParsedValue().orElse(false)) {
                final String value = stringOrEmpty(commandQueue.peek());
                commandContext.setCurrentArgument(child.getValue());
                return child.getValue().getSuggestionsProvider().apply(commandContext, value);
            }
            // END: Preprocessing

            // START: Parsing
            commandContext.setCurrentArgument(child.getValue());
            final ArgumentParseResult<?> result = child.getValue().getParser().parse(commandContext, commandQueue);
            if (result.getParsedValue().isPresent()) {
                commandContext.store(child.getValue().getName(), result.getParsedValue().get());
                return this.getSuggestions(commandContext, commandQueue, child);
            } else if (result.getFailure().isPresent()) {
                final String value = stringOrEmpty(commandQueue.peek());
                commandContext.setCurrentArgument(child.getValue());
                return child.getValue().getSuggestionsProvider().apply(commandContext, value);
            }
```

It is difficult to see the problem, until you start using suggestion providers. The problem is that when you add a StringArgument with a custom suggestion provider, the parsing will succeed even when the value being parsed isn't one of the suggestions. As a result, players can start typing and see the full list, but the moment they type one more character, all suggestions disappear.

I will probably fix this in my PR https://github.com/Incendo/cloud/pull/142 but need to know whether this was intentional behavior.

My proposed fix will be to not call the deeper getSuggestions() function when the number of remaining elements in the command queue is 0, and instead do the (parse) failure case.

The only workaround right now is to not use the suggestion provider, but make a custom argument parser which rejects values that aren't part of the enumeration. That's not a solution for people that want to accept non-suggested input.

bergerkiller:
Similar problem when theres command:
`/command <arg> literal`

And suggestions: [hello, helloworld]

[...]

PR #146, opened 2020-11-23
Fix partial command suggestions when using a suggestion provider

bergerkiller:
This addresses issue https://github.com/Incendo/cloud/issues/144 and possibly further completes https://github.com/Incendo/cloud/pull/122

Fixes suggestions (using a suggestion provider) not showing up while typing because the argument is parsed. Adds a new test which strictly evaluates all the possible behaviors.

Changed the way getSuggestions() works for dynamic arguments. When it tries to parse the command input, it then checks if there are other inputs remaining in the queue, and only then will it recursively go deeper into the command. Now a space needs to be typed in order to view the suggestions a level deeper.

This makes sure suggestion 'pineapple' shows up when the user types 'pine'.

After the parse attempt, it may have consumed the full command queue. To still have proper valid suggestions, the original command queue from before parsing is restored. Otherwise the command suggestion processor glitches out.

<details>
<summary>No longer relevant</summary>
**Important change**: This makes it impossible for suggestion providers/parsers to produce suggestions that do not match user input. For example, when the user types:
`/command he`

It can show suggestions 'help', 'hello' and 'hey', but 'usage' will not be a possible suggestion. If you try to include it in the suggestion provider anyway, it will be filtered as part of post-processing.

For similar reasons, it adds a (possibly) unneeded performance overhead because of filtering the suggestions. In some cases (such as NumberArgument) this is useless.

I've considered adding an extra property ('isSuggestionsFiltered()') to the ArgumentParser, however, I've noticed that everywhere a BiFunction is used for the suggestion provider. I would have to change that to an interface with a suggest() function to add it there. This would break existing API compatibility, so I have not included such changes here.

If changing BiFunction -> Interface and adding 'isSuggestionsFiltered() default true' function is desired, I will add that as well to this PR.
</details>

### DIFF
--- a/cloud-core/src/main/java/cloud/commandframework/CommandTree.java
+++ b/cloud-core/src/main/java/cloud/commandframework/CommandTree.java
@@ -449,87 +449,7 @@ public final class CommandTree<C> {
         }
         final List<Node<CommandArgument<C, ?>>> children = root.getChildren();
         if (children.size() == 1 && !(children.get(0).getValue() instanceof StaticArgument)) {
-            // The value has to be a variable
-            final Node<CommandArgument<C, ?>> child = children.get(0);
-
-            /* When we get in here, we need to treat compound arguments a little differently */
-            if (child.getValue() instanceof CompoundArgument) {
-                @SuppressWarnings("unchecked") final CompoundArgument<?, C, ?> compoundArgument = (CompoundArgument<?, C, ?>) child
-                        .getValue();
-                /* See how many arguments it requires */
-                final int requiredArguments = compoundArgument.getParserTuple().getSize();
-                /* Figure out whether we even need to care about this */
-                if (commandQueue.size() <= requiredArguments) {
-                    /* Attempt to pop as many arguments from the stack as possible */
-                    for (int i = 0; i < requiredArguments - 1 && commandQueue.size() > 1; i++) {
-                        commandQueue.remove();
-                        commandContext.store("__parsing_argument__", i + 2);
-                    }
-                }
-            } else if (child.getValue() instanceof FlagArgument) {
-                /* Remove all but last */
-                while (commandQueue.size() > 1) {
-                    commandContext.store(FlagArgument.FLAG_META, commandQueue.remove());
-                }
-            } else if (child.getValue() != null
-                    && GenericTypeReflector.erase(child.getValue().getValueType().getType()).isArray()) {
-                while (commandQueue.size() > 1) {
-                    commandQueue.remove();
-                }
-            } else if (child.getValue() != null
-                    && commandQueue.size() <= child.getValue().getParser().getRequestedArgumentCount()) {
-                for (int i = 0; i < child.getValue().getParser().getRequestedArgumentCount() - 1
-                        && commandQueue.size() > 1; i++) {
-                    commandContext.store(
-                            String.format("%s_%d", child.getValue().getName(), i),
-                            commandQueue.remove()
-                    );
-                }
-            }
-
-            if (child.getValue() != null) {
-                if (commandQueue.isEmpty()) {
-                    return Collections.emptyList();
-                } else if (child.isLeaf() && commandQueue.size() < 2) {
-                    commandContext.setCurrentArgument(child.getValue());
-                    return child.getValue().getSuggestionsProvider().apply(commandContext, commandQueue.peek());
-                } else if (child.isLeaf()) {
-                    if (child.getValue() instanceof CompoundArgument) {
-                        final String last = ((LinkedList<String>) commandQueue).getLast();
-                        commandContext.setCurrentArgument(child.getValue());
-                        return child.getValue().getSuggestionsProvider().apply(commandContext, last);
-                    }
-                    return Collections.emptyList();
-                } else if (commandQueue.peek().isEmpty()) {
-                    commandContext.setCurrentArgument(child.getValue());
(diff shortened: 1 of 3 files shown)


Proximyst (changes requested):
> Important change: This makes it impossible for suggestion providers/parsers to produce suggestions that do not match user input. For example, when the user types:
/command he

This is a HUGE no from me. What if I *want* to have that show e.g. `me`? Fuzzy matching through levenshtein distance would be crippled.

bergerkiller:
I agree, good example of a use case outside of prefix matching. Questions:
- Do you configure this filtering function using a register() function on the cloud manager (globally)?
- Do you configure this filtering function on the suggestion provider? Separate method with a default implementation?
- Should both be possible? Which one should override the other?

Say I want to use levenshtein distance to sort (and filter) results, would I as a user configure this globally? It might not make sense for things like numbers, but for most other types of name-based parsers this is desired.

All this is about is the three lines of code [here](https://github.com/Incendo/cloud/pull/146/commits/068aafb794e5aed033135ca1827ba17afac6f5b9#diff-1e41257f880f3e92a98419ac5886d28c0c39e074759aaeaea066cdbbc701501fR583).

What do you recommend?

Proximyst:
I'd do it on a suggestions provider, not for the entire command. Fuzzy matching would unlikely be very useful for static arguments; literals/static args should therefore behave as Brigadier would do it, but providers should really be given free way to do whatever they want.

We're not in the business of limiting users, though we may hold their hands at times. An option to disable *all* filtering would be welcome on the manager, but it should only ever affect static (i.e. not provided by a provider) arguments if any at all.

bergerkiller:
Alright, then suggest:
1. Change BiFunction to a new interface for suggestion providers
2. The ArgumentParser interface extends this same interface, that way, the same suggest() function (with same javadocs) is used
3. Add a default method to this interface, next to 'suggest', called 'suggestFiltered()' which has a default implementation that calls suggest() and performs the default prefix filtering behavior
4. Override suggestFiltered() in NumberParser (maybe others) to avoid unneeded overhead where the suggestions are already guaranteed to have the input as prefix

If you agree, I'll make these changes later today and update the PR

bergerkiller:
I have some changes open but I'm not happy with it. There are two conflicting systems at play here:
- Letting ArgumentParser/SuggestionProvider filter the suggestions based on input
- Letting the CommandManager getCommandSuggestionProcessor() post-filter the suggestions

I'm not sure how to proceed to make both happy. Right now the command suggestion processor is receiving already-filtered suggestions, effectively doing nothing. But I can imagine if I were to write my own suggestion filter I would want to see the unfiltered suggestions too.

I'm considering letting the implementation of CommandSuggestionProcessor call the actual filteredSuggestions() function on the SuggestionProvider. That way any custom processor can decide to call .suggestions() instead and filter that. The default implementation will just be a no-op that calls filteredSuggestions().

This way both this provider and the argument parsers themselves can perform filtering as they would like.

Thoughts?

bergerkiller:
Rewrite. No longer performs filtering in the Commandtree suggestions() function. Instead, all filtering is done in the Command Suggestion Processor. I've updated the processor so that it filters exactly-matching values. By setting the queue to the input, various bugs were fixed where the suggestion processor would fail.

jmanpenilla:
I am not a fan of lumping in changes to the default `CommandSuggestionProcessor` with this PR. Other than that, however, these changes look mostly good, and I've tested them to fix the issue I was having with #122.

If this PR were updated to only include https://github.com/Incendo/cloud/pull/146/commits/76272d18372db058d5e26da1c4a4c7788e4b2a82 (and not https://github.com/Incendo/cloud/pull/146/commits/6991ff93e1b08ff1e91486427501feed48ab340b), then it would get my approval.

bergerkiller:
I can do that. I'm not personally sure whether it should suggest the value the user has already typed, but it doesn't break anything keeping it in. Seems to be more a matter of taste, and could be overrided by setting your own suggestion processor anyway.

cloud-core/src/main/java/cloud/commandframework/CommandTree.java
```diff
@@ -449,87 +449,7 @@ private CommandTree(final @NonNull CommandManager<C> commandManager) {
...
-                    commandContext.store(child.getValue().getName(), result.getParsedValue().get());
-                    return this.getSuggestions(commandContext, commandQueue, child);
-                } else if (result.getFailure().isPresent()) {
-                    final String value = commandQueue.peek() == null ? "" : commandQueue.peek();
-                    commandContext.setCurrentArgument(child.getValue());
-                    return child.getValue().getSuggestionsProvider().apply(commandContext, value);
-                }
-                // END: Parsing
-            }
+            return suggestionsForDynamicArgument(commandContext, commandQueue,  children.get(0));
```
jmanpenilla, line 452:
```suggestion
            return this.suggestionsForDynamicArgument(commandContext, commandQueue,  children.get(0));
```

jmanpenilla (approved):
These changes look good to me. I've tested this PR to solve the issues I was having in #122, as well as various other related issues.

cloud-core/src/main/java/cloud/commandframework/CommandTree.java
bergerkiller: Changed that, and removed the double-space before children.get

Proximyst (approved):

PR #148, opened 2020-11-25
:sparkles: Give CommandManager a registration state

zml2008:
This replaces `LockableCommandManager` with something more universal -- since most platforms will use it somewhere.

The state tracking adds a pre-registration state, which restricts some of the command manager operations that change how future command registrations occur to before any commands have been registered.

I think I got most places where the state tracking makes sense, anything else stand out?

An open question: Should there be a way for users to bypass the registration limits, for situations where a registration too late might just result in completion issues (say, command trees have already been sent to clients)?


Sauilitired:
> An open question: Should there be a way for users to bypass the registration limits, for situations where a registration too late might just result in completion issues (say, command trees have already been sent to clients)?

Could make that a manager setting (https://github.com/Incendo/cloud/blob/a264605cf0d3c6e9a93a83ee4a70a856eb9daa68/cloud-core/src/main/java/cloud/commandframework/CommandManager.java#L777) so that people can disable the state checking if they really want to

zml2008:
will look at tomorrow, maybe call it `ALLOW_UNSAFE_REGISTRATION`?

Sauilitired:
> will look at tomorrow, maybe call it `ALLOW_UNSAFE_REGISTRATION`?

That sounds good :+1:

zml2008:
done :)

Sauilitired:
bruh wtf am i dumb that's no emoji

PR #153, opened 2020-11-30
Rename CommandContext getOrDefault to getorSupplyDefault

bergerkiller:
A method was added to CommandContext to also allow a supplier to supply a value. However, this caused an ambiguous function situation when called with e.g. _null_. The other getOrDefault(key, value) allows for a nullable value, but when the supplier is null, an NPE is thrown.

Up to you whether you find this important enough of an issue. I just had a dependency of cloud suddenly throwing errors only to find out about this change :)

Workaround fix is doing `getOrDefault(key, (Object) null)`

### DIFF
--- a/cloud-core/src/main/java/cloud/commandframework/context/CommandContext.java
+++ b/cloud-core/src/main/java/cloud/commandframework/context/CommandContext.java
@@ -248,7 +248,7 @@ public final class CommandContext<C> {
      * @return Argument, or supplied default value
      * @since 1.2.0
      */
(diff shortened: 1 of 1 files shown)


jmanpenilla:
> I just had a dependency of cloud suddenly throwing errors only to find out about this change :)

To be fair, you are using snapshot builds which we currently don't publish anywhere (for reasons like this :P)
Besides that though, if this made it to a release it would be API breaking, so thanks for the fix :)

jmanpenilla (approved):

PR #174, opened 2020-12-16
Fabric support

zml2008:
Re-opening based on the previous PR

This is a basic implementation that can register commands, but doesn't have any fancy argument types. I'm hoping to add that eventually. I'll probably go about that by allowing argument parsers to delegate to Brigadier types, since that will be easier, and allow for easier integration with mods that provide their own Brigadier argument types.

# Argument type support

This PR aims to implement all Vanilla argument types (besides the development-only ones), primarily through wrapping the Brigadier parsers. Current progress:

## Exist in Cloud

- [x] `UUIDArgumentType` (network: `minecraft:uuid`)

## Directly mapped

- [x] `ColorArgument` (network: `minecraft:color`)
- [x] `CompoundTagArgument` (network: `minecraft:nbt_compound_tag`)
- [x] `NbtTagArgument` (network: `minecraft:nbt_tag`)
- [x] `NbtPathArgument` (network: `minecraft:nbt_path`) **todo:** the actual implementation for this is an inner class of the brigadier type, does it make sense to wrap that?
- [x] `ScoreboardCriteriaArgument` (network: `minecraft:objective_criteria`)
- [x] `ScoreboardOperationArgument` (network: `minecraft:operation`)
- [x] `ParticleEffectArgument` (network: `minecraft:particle`)
- [x] `AngleArgument` (network: `minecraft:angle`) **todo:** the vanilla impl is logical server-only, but doesn't need to be 
- [x] `AxisArgument` (network: `minecraft:swizzle`)
- [x] `IdentifierArgument` (network: `minecraft:resource_location`)
- [x] `EntityAnchorArgument` (network: `minecraft:entity_anchor`)
- [x] `IntRangeArgument` (network: `minecraft:int_range`)
- [x] `FloatRangeArgument` (network: `minecraft:float_range`)
- [x] `ItemStackArgument` (network: `minecraft:item_stack`) **todo:** this might want a wrapper -- it's not a full item stack, just the item type and NBT tag

## Need wrapper type

These Brigadier parsers only return an intermediate type, and require a ServerCommandSource to be fully resolved. We need to evaluate which ones actually need a server, which can be made client-friendly. I expect everything that uses selectors will have to be server-only, but others can be client-friendly

These would also be bound to a ServerCommandSource sender type, or would need to provide a mapper function.

### Using selectors
- [ ] `ScoreHolderArgument` (network: `minecraft:score_holder`) -- single vs multiple, anything that can be on a scoreboard (all wildcarding is done based on selector matches)
- [ ] `EntityArgument` (network: `minecraft:entity`) -- supports entity/player, and single/multiple properties
- [x] `MessageArgument` (network: `minecraft:message`) -- this is just a greedy string that resolves selectors
- [ ] `GameProfileArgument` (network: `minecraft:game_profile`) - parses selectors

### Using relative positions
- [ ] `BlockPosArgument` (network: `minecraft:block_pos`) - can be relative
- [ ] `ColumnPosArgument` (network: `minecraft:column_pos`) - can be relative
- [ ] `Vec3dArgument` (network: `minecraft:vec3`) - can be relative
- [ ] `Vec2fArgument` (network: `minecraft:vec2`) - can be relative

### Others/to be sorted

- [ ] `BlockPredicateArgument` (network: `minecraft:block_predicate`)
- [ ] `BlockStateArgument` (network: `minecraft:block_state`)
- [ ] `ItemPredicateArgument` (network: `minecraft:item_predicate`)
- [ ] `ScoreboardObjectiveArgument` (network: `minecraft:objective`)
- [ ] `RotationArgument` (network: `minecraft:rotation`)
- [ ] `ScoreboardSlotArgument` (network: `minecraft:scoreboard_slot`)
- [x] `TeamArgument` (network: `minecraft:team`)
- [ ] `ItemSlotArgument` (network: `minecraft:item_slot`)
- [ ] `CommandFunctionArgument` (network: `minecraft:function`)
- [x] `TimeArgument` (network: `minecraft:time`) -- time in ticks

## Combined to one registry values argument

These are all specialized argument types that can be exposed as one single argument type in-API. We'll be able to send the appropriate argument type/suggestion provider to be used for completions though, falling back to querying the server.

- [x] `EntityTypeArgument` (network: `minecraft:entity_summon`)
- [x] `AVAILABLE_SOUNDS` suggestion provider
- [x] `ALL_BIOMES` suggestion provider
- [x] `SUMMONABLE_ENTITIES` (all entities but player and fishing bobber)
- [x] `EnchantmentArgument` (network: `minecraft:item_enchantment`) (needs to be re-implemented)
- [x] `ServerWorldArgument` (network: `minecraft:dimension`)
- [x] `StatusEffectArgument` (network: `minecraft:mob_effect`) (needs to be re-implemented)

## Doesn't exist as an argument type, but is a client-local completion provider

- [ ] `ALL_RECIPES`

## Entirely custom

Some parsers that might be useful for mod developers but aren't included in base Cloud:

- [x] `RegistryEntryArgument` -- any entry in a registry, should support both static and dynamic registries

# Other odds and ends

- [ ] Register mapped argument parsers with Brigadier
- [ ] Add support for split server/sent to client command trees, so the server doesn't double-parse
- [ ] Integrate with client commands api once FabricMC/fabric#1115 is merged


cloud-minecraft/cloud-fabric/src/main/resources/fabric.mod.json
```diff
@@ -0,0 +1,21 @@
+{
+  "schemaVersion": 1,
+  "id": "cloud",
+  "version": "${version}",
+
+  "name": "Cloud",
+  "description": "Command framework and dispatcher for the JVM",
+  "authors": [ "Alexander Söderberg" ],
```
Citymonstret, line 8: You should probably add yourself here? We can switch to GitHub usernames :eyes:

Issue #160, opened 2020-12-08
Discussion: Support multiple variables


Sauilitired:
**Context:**
- https://github.com/Incendo/cloud/issues/137
- https://github.com/Incendo/cloud/pull/142

**Description:**
Cloud currently supports having multiple string literals OR a single variable the children of any parent node. This means that  Cloud allows `/command literal1|literal2|literal3` and `/command <argument>` but not `/command literal|<argument>` or `/command <argument1>|<argument2>`

This was done to guarantee that no ambiguous cases can exist in a command tree, which means that we get completely deterministic input evaluation.

This is however a quite major limitation and this decision has been somewhat controversial. The decision is not final and I think it's about time we come to a conclusion.

**Possible Solutions:**

1. We can do nothing and keep the current limitations.

2. We can support multiple literals and a single argument. This has been attempted in pull request #142. This requires that all string literals are parsed before any variable arguments and this restriction should be appended to the library contracts.

3. We can support multiple literals and multiple arguments. This would be the less restrictive path, but has some major problems: how do we handle error feedback when invalid command input is provided? what order do we evaluate the arguments in? how do we deal with platform integrations (most notably, Brigadier)?

The first option is obviously the easiest but appears to be less favourable. The second option is possible as has been demonstrated by @bergerkiller. The third option would be the least restrictive but would probably require a major rework of the library internals. 

All opinions are very welcome!

SirYwell:
I'd prefer the second option.

The limitations to avoid ambiguousity make sense, but checking literals before passing input as an argument is neither a performance issue nor does it depend on any ordering other than literals before argument. When implementing a command of that type, one need to make sure that e.g. all player names still can be used tho.

Having multiple arguments, however, requires a well-defined order of arguments to check and as already mentioned, error handling is way more difficult and such usage would likely have a negative impact on user experience.

Sauilitired:
The people discussing it has largely favoured the second solution. However, some concerns have been brought up regarding this solution and how pattern matching must necessarily introduce additional complexity. The third solution can most likely be removed from the equation.

bergerkiller:
The major problem with the second option (which I chose) is that you have to make sure that the argument portion is never expected to be one of the literals. For instance, if you have a <name> argument with literals overriding it, it means that <name> can never equal one of the literals.

A way around that would be to offer some form of syntax to 'escape' a literal. For example, you could make it so that:
```
/command literal - executes literal
/command "literal" - executes <arg> with "literal" as input
```

[...]

PR #178, opened 2020-12-17
Re-implement command tree allowing literal arguments to override dynamic ones

bergerkiller:
Implements suggestion number 2 from https://github.com/Incendo/cloud/issues/160

### DIFF
--- a/cloud-core/src/main/java/cloud/commandframework/CommandTree.java
+++ b/cloud-core/src/main/java/cloud/commandframework/CommandTree.java
@@ -49,13 +49,16 @@ import java.util.Collection;
 import java.util.Collections;
 import java.util.Comparator;
 import java.util.HashMap;
+import java.util.HashSet;
 import java.util.Iterator;
 import java.util.LinkedList;
 import java.util.List;
 import java.util.Map;
 import java.util.Objects;
 import java.util.Queue;
+import java.util.Set;
 import java.util.stream.Collectors;
+import java.util.stream.Stream;
 
 /**
  * Tree containing all commands and command paths.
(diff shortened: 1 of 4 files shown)


cloud-minecraft/cloud-brigadier/src/main/java/cloud/commandframework/brigadier/CloudBrigadierManager.java
```diff
@@ -574,6 +593,12 @@ public void registerDefaultArgumentTypeSupplier(
         }
 
         for (final String suggestion : suggestions) {
+            // Skip suggestions that are sibling literals
+            // They are already part of the literal argument construction earlier
+            if (siblingLiterals.contains(suggestion)) {
```
Sauilitired, line 598: This can be done before the loop using `suggestions.removeIf(siblingLiterals::contains)`. While this would require an extra loop, I would prefer keeping that logic outside of the "rendering" loop

cloud-core/src/main/java/cloud/commandframework/CommandTree.java
```diff
@@ -268,157 +271,182 @@ private CommandTree(final @NonNull CommandManager<C> commandManager) {
...
-                            Node<CommandArgument<C, ?>> node = child;
-                            while (!node.isLeaf()) {
-                                node = node.getChildren().get(0);
-                                if (node.getValue() != null && node.getValue().getOwningCommand() != null) {
-                                    child.getValue().setOwningCommand(node.getValue().getOwningCommand());
+
+        // Check whether it matches any of the static arguments
+        // If so, do not attempt parsing as a dynamic argument
+        boolean matchesLiteral = false;
+        if (!commandQueue.isEmpty()) {
```
Sauilitired, line 278: Can't we return immediately if we know it's matching a literal? That would get rid of the if statement below

cloud-minecraft/cloud-brigadier/src/main/java/cloud/commandframework/brigadier/CloudBrigadierManager.java
bergerkiller: I don't think the list returned by suggest() is modifiable like that, but I could stream(), filter() and collect() it into a new List

cloud-core/src/main/java/cloud/commandframework/CommandTree.java
bergerkiller: Good catch, must've left that in there from earlier. I'll fix that.

bergerkiller:
For your 'keeping logic outside of the loop' I've changed it so it fills a new ArrayList and then removes elements from it

Sauilitired (approved):

cloud-minecraft/cloud-brigadier/src/main/java/cloud/commandframework/brigadier/CloudBrigadierManager.java
```diff
@@ -497,6 +503,18 @@ public void registerDefaultArgumentTypeSupplier(
                             )))
                     .executes(executor);
         } else {
+            // Check for sibling literals (StaticArgument)
+            // These are important when providing suggestions
+            final Set<String> siblingLiterals = (root.getParent() == null) ? Collections.emptySet()
+                    : root.getParent().getChildren().stream()
+                        .filter(n -> n.getValue() instanceof StaticArgument)
+                        .map(n -> (StaticArgument<C>) ((CommandArgument) n.getValue()))
+                        .flatMap(s -> s.getAliases().stream())
```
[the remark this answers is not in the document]
Sauilitired: Yeah, might as well keep it in the case that we change it (possibly adding a toggle for it)

Sauilitired:
In the future, please make sure to tick the box that let's us edit the PR. I created a new PR based off of your PR that we can edit.

PR #195, opened 2021-01-01
✨ Convert build scripts to Kotlin

jmanpenilla:
Also added some new tasks to only build/install to maven local certain platforms
`buildMinecraft`, `installMinecraft`, `buildDiscord`, `installDiscord`, `buildIRC`, `installIRC`


buildSrc/src/main/kotlin/DependencyVersions.kt
```diff
@@ -0,0 +1,29 @@
...
+        "jb-annotations" to "20.1.0",
+        "guava" to "21.0-jre",
+        // IRC DEPENDENCIES
+        "pircbotx" to "83a4c22e80",
+        // TEST DEPENDENCIES
+        "jupiter-engine" to "5.7.0",
+        "jhm" to "1.25.2"
+)
+
+val vers
```
Sauilitired, line 28: What's the point of this?

build.gradle.kts
```diff
@@ -0,0 +1,191 @@
...
+
+checkstyle {
+    configFile = file("config/checkstyle/checkstyle.xml")
+}
+
+buildGroups("Minecraft", "Discord", "IRC")
+
+gradle.taskGraph.whenReady {
+    gradle.taskGraph.allTasks.forEach {
+        if (it.project.name.contains("example")) {
```
Sauilitired, line 24: Should we create a group for the examples as well, to be consistent?

Sauilitired (commented):
This looks good :D

buildSrc/src/main/kotlin/DependencyVersions.kt
jmanpenilla: the map was called `vers` in the old scripts so it was convenient to have, I can just use the ide to rename it now though

build.gradle.kts
jmanpenilla: the example are a bit different as we *only* want to build them when we include the flag, as opposed to the groups where normally you build everything but the group allows building a smaller subsection.

Sauilitired: fair

buildSrc/src/main/kotlin/DependencyVersions.kt
```diff
@@ -0,0 +1,29 @@
+val versions = mapOf(
```
zml2008, line 1: what people usually do for this is something like:

```kotlin
object DependencyVersions {
  val checkerQual = "3.8.0"
  val geantyref = "1.3.11
}
```

which gives you some extra type-safety and IDE support

jmanpenilla: I also thought about some extension functions to bring in dependencies like DependencyHandlerScope.checkerQual() but maybe that is going too far

zml2008: I've done that sometimes, but only really for dependencies that need to exist in multiple configurations at once -- often AP + compileOnlyApi annotation lib combinations

PR #198, opened 2021-01-04
Register JDA parsers

p5nbTgip0r:
This PR registers the `MessageChannel` and `User` argument types for JDA. I also changed a few `List` types to `Set` for storing the active parser modes in `ChannelArgument.java` to match `UserArgument.java`


broccolai (approved):
LGTM, thank you!

PR #199, opened 2021-01-05
JDA argument parser fixes

p5nbTgip0r:
## Fix mentioned user argument parsing if a mention is not given
If only `ParserMode.MENTION` was set as a parsing mode, the initial check for `input.endsWith(">")` would be bypassed and the input would begin to be parsed like a mention. If the input was less than three characters long then the `substring` calls would give a very ambiguous error along the lines of `begin 2, end X-1, length X`.

This PR fixes this issue by adding a check to make sure the input is an actual user mention. If the input isn't a mention then an exception will be set stating that the sender's input was not a user mention.

In addition, this PR also removes the `modes.size() == 1` check in the mention parser. I'm not quite sure why this was here, but I don't think that the input parsing check should be bypassed because only one parsing mode is specified. I settled on a check in the `UserParser` constructor to make sure that the parsing modes list isn't empty in order to ensure that at least one parser mode handles the input.

## Fix channel argument parsing
This part has similar changes to the user parser (in regards to making sure at least one parser mode is set and removing the `|| modes.size == 1` check), but it also has some important changes in it as well:

Because the channel argument parser relies on having access to the JDA guild instance, I added a check to make sure that the command context contains `MessageReceivedEvent`. In addition, I also added a check to make sure that the command was not issued from a DM because channels are only available in a guild.

## Add more context keys to JDA preprocessor
This portion adds a vital key (`MessageReceivedEvent`) and a few useful keys to be stored in the command context. Channel argument parsing relies on having `MessageReceivedEvent` available in the command context.

## Adjust JDA example to provide `MessageReceivedEvent` in commandSender mappers
Adding the `MessageReceivedEvent` to the command context requires `backwardsCommandSenderMapper` to be able to map a custom command sender into a `MessageReceivedEvent` successfully. If the [`event` parameter of `JDACommandSender`](https://github.com/Incendo/cloud/blob/547ecf98ffc81741644b0ddda4f013878ed2ce49/cloud-discord/cloud-jda/src/main/java/cloud/commandframework/jda/JDACommandSender.java#L53) is null then the [`JDA4CommandManager`'s `backwardsCommandSenderMapper` will throw a `IllegalStateException`](https://github.com/Incendo/cloud/blob/547ecf98ffc81741644b0ddda4f013878ed2ce49/cloud-discord/cloud-jda/src/main/java/cloud/commandframework/jda/JDA4CommandManager.java#L77). To avoid this, the `MessageReceivedEvent` must be stored in the custom command sender objects and then passed back into the `JDACommandSender` constructor when needed in the `backwardsCommandSenderMapper`.

### DIFF
--- a/cloud-discord/cloud-jda/src/main/java/cloud/commandframework/jda/JDACommandPreprocessor.java
+++ b/cloud-discord/cloud-jda/src/main/java/cloud/commandframework/jda/JDACommandPreprocessor.java
@@ -25,6 +25,9 @@ package cloud.commandframework.jda;
 
 import cloud.commandframework.execution.preprocessor.CommandPreprocessingContext;
 import cloud.commandframework.execution.preprocessor.CommandPreprocessor;
+import net.dv8tion.jda.api.entities.ChannelType;
+import net.dv8tion.jda.api.entities.Guild;
+import net.dv8tion.jda.api.events.message.MessageReceivedEvent;
 import org.checkerframework.checker.nullness.qual.NonNull;
 
 /**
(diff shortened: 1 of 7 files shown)


broccolai (approved):
LGTM, thanks!

PR #200, opened 2021-01-08
Expand ability to use Adventure Components in exceptions and metadata

zml2008:
This makes the MinecraftExceptionHandler properly handle exception causes that are `ComponentMessageThrowable`s.

I've added `Component`-based metadata here that will be used in `MinecraftHelp`. They don't currently get written to the standard description keys unfortunately, but we might be able to use #149 to transform command meta to resolve that.

(I've also tweaked the Components in MinecraftExceptionHandler to have a slightly more effecient structure)

### DIFF
--- a/cloud-minecraft/cloud-minecraft-extras/src/main/java/cloud/commandframework/minecraft/extras/MinecraftExtrasMetaKeys.java
+++ b/cloud-minecraft/cloud-minecraft-extras/src/main/java/cloud/commandframework/minecraft/extras/MinecraftExtrasMetaKeys.java
@@ -37,7 +37,7 @@ public final class MinecraftExtrasMetaKeys {
     /**
      * A component short description.
      *
-     * <p>This will set the plain-text description, but will be used in place of that meta key in help.</p>
+     * <p>This will not set the plain-text description, but will be used in place of that meta key in help.</p>
      */
     public static final CommandMeta.Key<Component> DESCRIPTION = CommandMeta.Key.of(
             Component.class,
@@ -47,7 +47,7 @@ public final class MinecraftExtrasMetaKeys {
     /**
      * A component long description.
      *
-     * <p>This will set the plain-text description, but will be used in place of that meta key in help.</p>
+     * <p>This will not set the plain-text long description, but will be used in place of that meta key in help.</p>
      */
     public static final CommandMeta.Key<Component> LONG_DESCRIPTION = CommandMeta.Key.of(
             Component.class,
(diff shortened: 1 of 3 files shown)


cloud-minecraft/cloud-minecraft-extras/src/main/java/cloud/commandframework/minecraft/extras/MinecraftExtrasMetaKeys.java
```diff
@@ -0,0 +1,60 @@
...
+ * Extra command metadata for providing rich text.
+ *
+ * @since 1.4.0
+ */
+public final class MinecraftExtrasMetaKeys {
+
+    /**
+     * A component short description.
+     *
+     * <p>This will set the plain-text description, but will be used in place of that meta key in help.</p>
```
jmanpenilla, line 40: will or will not?

cloud-minecraft/cloud-minecraft-extras/src/main/java/cloud/commandframework/minecraft/extras/MinecraftExtrasMetaKeys.java
```diff
@@ -0,0 +1,60 @@
...
+     */
+    public static final CommandMeta.Key<Component> DESCRIPTION = CommandMeta.Key.of(
+            Component.class,
+            "cloud:minecraft_extras/description"
+    );
+
+    /**
+     * A component long description.
+     *
+     * <p>This will set the plain-text description, but will be used in place of that meta key in help.</p>
```
jmanpenilla, line 50: will or will not?

cloud-minecraft/cloud-minecraft-extras/src/main/java/cloud/commandframework/minecraft/extras/MinecraftHelp.java
```diff
@@ -439,9 +447,21 @@ private void printVerboseHelpTopic(
...
-                ? this.messageProvider.provide(sender, MESSAGE_NO_DESCRIPTION)
-                : this.descriptionDecorator.apply(helpTopic.getDescription());
+        /* Topics will use the long description if available, but fall back to the short description. */
+        final Component richDescription =
+                helpTopic.getCommand().getCommandMeta().get(MinecraftExtrasMetaKeys.LONG_DESCRIPTION)
+                        .orElse(helpTopic.getCommand().getCommandMeta().get(MinecraftExtrasMetaKeys.DESCRIPTION)
+                                .orElse(Component.empty()));
+
+        final Component topicDescription;
+        if (richDescription != Component.empty()) {
```
jmanpenilla, line 457: Is there any benefit to using `Component.empty()` instead of just `null` here?

zml2008: uh not really, it's just habit to use the empty component rather than null since that's what we do in a lot of places in Adventure.

PR #203, opened 2021-01-09
Use type-safe sender-type casts

Proximyst:
This PR takes the goal of adding type-safe sender type casts such that following code can be written:

```java
var manager = new ExampleCommandManager<Supertype>();
var command = manager.commandBuilder("example")
    .senderType(Subtype.class)
    .handler(ctx -> {
        ctx.getSender(); // Subtype
    });
manager.command(command);
```

This is done through some generic changes and a couple casts that should be safe given the new generic type `N` holds the expectation `N extends C`, where `C` is the old type. This is safe, because by the time the command is called, the sender type should already be checked. `N` is therefore always going to be assignable to the type `C` (like how `String` is assignable to `CharSequence`, and `(Class<CharSequence>) classInstance` where `Class<? extends CharSequence> classInstance = String.class` is safe).

---

This should AFAIK be ABI-compatible with the previous code.


broccolai (approved):

JarFiles (approved):

zml2008 (approved):

broccolai:
closing for #518

PR #204, opened 2021-01-11
Expose Command/Context from exceptions where possible

jmanpenilla:
This exposes the `Command` which caused the `InvalidCommandSenderException`, which can be useful for changing exception handling based on command meta, and also exposes the `CommandContext` leading up to `CommandExecutionException`s.


broccolai (approved):
LGTM, just needs to be added to the changelog

PR #212, opened 2021-01-13
sponge7: Implement basic command manager for SpongeAPI 7

zml2008:
A simple implementation of a command manager on SpongeAPI.

Since v7 is rapidly becoming legacy software, I haven't spent time implementing any argument types -- I expect this will be used mostly by plugins that are already multiplatform. If requested, those can be added later of course.


cloud-minecraft/cloud-sponge7/src/main/java/cloud/commandframework/sponge7/CloudCommandCallable.java
```diff
@@ -0,0 +1,227 @@
...
+        if (possible.isPresent()) {
+            return possible;
+        }
+
+        return this.getShortDescription(source);
+    }
+
+    @Override
+    public Text getUsage(final @NonNull CommandSource source) {
+        return Text.of(this.manager.getCommandSyntaxFormatter().apply(
```
zml2008, line 221: Is this correct? not sure

Citymonstret: https://github.com/Incendo/cloud/blob/e73c9a3bafe66a4c3846be57825ae4e8ef2ed868/cloud-core/src/main/java/cloud/commandframework/arguments/StandardCommandSyntaxFormatter.java#L53

I should probably clarify it in the JavaDoc. Basically, the list of arguments is meant to indicate what has already been supplied, i.e. the path that has taken you to the current command, whereas the nodes following that will explain what arguments can come *after* the current node.

Thus, I suppose the argument list should be empty (or the root), as nothing has been entered at this point.

zml2008: Ah, that makes sense. I've updated to use an empty list.

PR #214, opened 2021-01-15
JDA: Fix webhook messages throwing NPE

p5nbTgip0r:
Webhook messages would be constructed as `JDAGuildSender`, which calls `Objects.requireNonNull(event.getMember())` however the JDA documentation states that the `getMember()` method will "be null in case of Message being received in a PrivateChannel or isWebhookMessage() returning true." This PR makes it so if the message is a webhook message, a generic `JDACommandSender` instance will be constructed instead of a `JDAGuildSender` which doesn't require calling `getMember()`


Citymonstret:
Can you add this change to the CHANGELOG? Otherwise it looks good to me!

p5nbTgip0r:
> Can you add this change to the CHANGELOG?

Sure, I've added it in the newest commit

PR #220, opened 2021-01-26
JDA: Add isolation levels to user argument

p5nbTgip0r:
Currently, when a user argument is parsed, cloud will use the `JDA` class's `getUsersByName()`/`getUserById()` method which attempts to fetch a user from anywhere that the bot can see. This means that if a user provides a name or ID for a user that is in a different guild, the bot may leak information about users in a different guild if the command handler does not check if the user is in the same guild that the command message was sent from. 

This PR adds two levels of isolation to this argument:
`GLOBAL` = the current functionality (any user the bot can see).
`GUILD` = only members in the same guild that the command was executed in.

Using the `GUILD` isolation with a command that is executed in a DM will only allow the author to specify themself for the user argument.

### DIFF
--- a/cloud-discord/cloud-jda/src/main/java/cloud/commandframework/jda/parsers/UserArgument.java
+++ b/cloud-discord/cloud-jda/src/main/java/cloud/commandframework/jda/parsers/UserArgument.java
@@ -121,7 +121,7 @@ public final class UserArgument<C> extends CommandArgument<C, User> {
     public static final class Builder<C> extends CommandArgument.Builder<C, User> {
 
         private Set<ParserMode> modes = new HashSet<>();
-        private Isolation isolationLevel = Isolation.GUILD;
+        private Isolation isolationLevel = Isolation.GLOBAL;
 
         private Builder(final @NonNull String name) {
             super(User.class, name);
--- a/cloud-discord/cloud-jda/src/main/java/cloud/commandframework/jda/parsers/UserArgument.java
+++ b/cloud-discord/cloud-jda/src/main/java/cloud/commandframework/jda/parsers/UserArgument.java
@@ -28,15 +28,18 @@ import cloud.commandframework.arguments.parser.ArgumentParseResult;
 import cloud.commandframework.arguments.parser.ArgumentParser;
 import cloud.commandframework.context.CommandContext;
 import cloud.commandframework.exceptions.parsing.NoInputProvidedException;
-import net.dv8tion.jda.api.JDA;
+import net.dv8tion.jda.api.entities.Member;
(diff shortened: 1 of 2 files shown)


cloud-discord/cloud-jda/src/main/java/cloud/commandframework/jda/parsers/UserArgument.java
```diff
@@ -153,19 +176,22 @@ private Builder(final @NonNull String name) {
...
         /**
          * Construct a new argument parser for {@link User}
          *
          * @param modes List of parsing modes to use when parsing
-         * @throws java.lang.IllegalStateException If no parsing modes were provided
+         * @param isolationLevel Level of isolation to maintain when parsing
+         * @throws java.lang.IllegalArgumentException If no parsing modes were provided
          */
-        public UserParser(final @NonNull Set<ParserMode> modes) {
+        public UserParser(final @NonNull Set<ParserMode> modes, final @NonNull Isolation isolationLevel) {
```
broccolai, line 188: You need to depreciate the old constructor and have it redirect to this new one

cloud-discord/cloud-jda/src/main/java/cloud/commandframework/jda/parsers/UserArgument.java
```diff
@@ -106,10 +112,16 @@ private UserArgument(
...
+    public enum Isolation {
+        GLOBAL,
+        GUILD
+    }
+
 
     public static final class Builder<C> extends CommandArgument.Builder<C, User> {
 
         private Set<ParserMode> modes = new HashSet<>();
+        private Isolation isolationLevel = Isolation.GUILD;
```
broccolai, line 124: I think the default, at least until 2.0 should be GLOBAL so we're not changing behaviour on a minor version

broccolai (changes requested):
Looks good! Just a few minor things

cloud-discord/cloud-jda/src/main/java/cloud/commandframework/jda/parsers/UserArgument.java
p5nbTgip0r: My thinking was that users' privacy could be impacted by GLOBAL and so GUILD would be a better default option, but I understand your concern. I've changed the default to be GLOBAL in my newest changes.

broccolai: Yeah, I do agree. We can change the default for 2.0

broccolai (approved):

PR #227, opened 2021-02-07
fixes for '/execute'

jpenilla:


zml2008 (approved):
This looks nasty but as long as it seems to work it's probably the best way to go for now

jpenilla:
> This looks nasty but as long as it seems to work it's probably the best way to go for now

Yeah it is pretty nasty but it does seem to work

PR #228, opened 2021-02-10
Map more arguments from NMS

MrIvanPlays:
I plan on porting the ItemStack argument as well as others I can find to port easily.

### DIFF
--- a/cloud-minecraft/cloud-bukkit/src/main/java/cloud/commandframework/bukkit/parsers/item/ItemStackParser.java
+++ b/cloud-minecraft/cloud-bukkit/src/main/java/cloud/commandframework/bukkit/parsers/item/ItemStackParser.java
@@ -49,18 +49,36 @@ public final class ItemStackParser<C> implements ArgumentParser<C, ItemStack> {
             @NonNull final CommandContext<@NonNull C> commandContext,
             @NonNull final Queue<@NonNull String> inputQueue
     ) {
-        // todo: make it possible to have `minecraft:diamond{display:{Name:{"text":{"The diamond award"}}}}
-        // currently it is possible to have NBT after the item specified but it is not possible to have a multi
-        // word message. I don't know how to so please someone who is more experienced than me in creating
-        // cloud arguments please tell me how to.
-        String input = inputQueue.peek();
-        if (input == null) {
+        String start = inputQueue.peek();
+        if (start == null) {
             return ArgumentParseResult.failure(new NoInputProvidedException(
                     ItemStackParser.class,
                     commandContext
             ));
         }
         inputQueue.remove();
+        StringBuilder inputBuilder = new StringBuilder(start);
+        int queueSize = inputQueue.size();
+        if (queueSize != 0 && start.indexOf('{') != -1) {
+            for (int i = 0; i < queueSize; i++) {
+                String peek = inputQueue.peek();
+                if (peek == null) {
+                    inputQueue.remove();
+                    continue;
+                }
+                if (i == 0) {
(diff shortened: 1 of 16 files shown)


zml2008:
We should look at maybe breaking the types used for Fabric wrapped arguments out into a common API module -- some of those could be used here.

ItemStack is complicated as well, unfortunately. The Mojang argument type is `<item registry value>[optional NBT TAG]`, and followed by a `count` argument. Probably the only effective way to build an item stack in the expected way would be to use a compound argument type of (ProtoItemStack, int)`

MrIvanPlays:
@zml2008 I don't know how this will turn out... I'm planning to use adventure-nbt for cloud-bukkit and you can use plain nbt from fabric. What can we abstract to be used in both the fabric implementation and the bukkit implementation?

cloud-minecraft/cloud-bukkit/src/main/java/cloud/commandframework/bukkit/parsers/item/ItemStackParser.java
```diff
@@ -0,0 +1,83 @@
...
+
+    @Override
+    public @NonNull ArgumentParseResult<@NonNull ItemStack> parse(
+            @NonNull final CommandContext<@NonNull C> commandContext,
+            @NonNull final Queue<@NonNull String> inputQueue
+    ) {
+        // todo: make it possible to have `minecraft:diamond{display:{Name:{"text":{"The diamond award"}}}}
+        // currently it is possible to have NBT after the item specified but it is not possible to have a multi
+        // word message. I don't know how to so please someone who is more experienced than me in creating
+        // cloud arguments please tell me how to.
```
MrIvanPlays, line 55: @Citymonstret ?

MrIvanPlays:
ItemStack argument
1.12.2 ( implementation for 1.8 - 1.12.2 ):


1.16.5 ( implementation for 1.13 - 1.16.5, brigadier support enabled using commodore ):


Test command:
```java
      commandManager.command(
          commandManager
              .commandBuilder("test")
              .argument(ItemStackArgument.<CommandSender>newBuilder("item").asRequired().build())
              .handler(
                  handle -> {
                    CommandSender sender = handle.getSender();
                    ItemStack item = handle.get("item");
                    sender.sendMessage("Type: " + item.getType());
                    ItemMeta meta = item.getItemMeta();
                    if (meta.hasDisplayName()) {
                      sender.sendMessage("Display name: " + meta.getDisplayName());
                    }
                    if (meta.hasLocalizedName()) {
                      sender.sendMessage("Localized name: " + meta.getLocalizedName());
                    }
                    if (meta.hasLore()) {
                      sender.sendMessage("Lore: " + Joiner.on(", ").join(meta.getLore()));
                    }
                  }));
```

MrIvanPlays:
um i am still alive. 
can i get a review on that please

MrIvanPlays:
bump

Proximyst:
Regarding the ItemStack stuff, I would rather be inclined to suggest a new API to Paper for parsing *current version* item stacks. For older versions, you will still need some ugly hacks, but would not be an issue for same version.

Give me a heads up if you want me to look into it for you, or you need help. There's simply little reason to handle parsing manually when the server holds all the cards already.

jpenilla:
For modern versions, I took a different approach of wrapping the Brigadier ArgumentType using WrappedBrigadierParser in #257. This works nicely for modern versions, but it's not that useful on pre-1.13 versions. I decided to just fall back to parsing Material for legacy versions.

Feel free to open separate PRs with your improvements to the legacy `ProtoItemStack` parser and your Time argument.

PR #237, opened 2021-03-14
Let ArgumentDescription#getDescription and #isEmpty have command sender

egg82:
Currently, ArgumentDescriptions have no way to tell who is requesting the description.

This is a useful feature for sending different descriptions depending on the person requesting them.
For example, a custom localization manager would require the requesting user to be known (for platforms without Adventure)

This PR generifies `ArgumentDescription` and `Description` and adds a `C sender` argument to the `getDescription` and `isEmpty` methods.

Potential downsides of this would be the lack of backwards compatibility, but because the signatures have changed compiler errors will be issued rather than runtime exceptions.

Seems to compile fine with `./gradlew clean build`

### DIFF
--- a/cloud-core/src/main/java/cloud/commandframework/ArgumentDescription.java
+++ b/cloud-core/src/main/java/cloud/commandframework/ArgumentDescription.java
@@ -32,49 +32,54 @@ import static java.util.Objects.requireNonNull;
 /**
  * A description for a {@link CommandArgument}
  *
+ * @param <C> Command sender type
  * @since 1.4.0
  */
-public interface ArgumentDescription {
+public interface ArgumentDescription<C> {
 
     /**
      * Get an empty command description.
      *
+     * @param <C> Command sender type
      * @return Command description
      */
     @SuppressWarnings("deprecation")
-    static @NonNull ArgumentDescription empty() {
-        return Description.EMPTY;
+    static <C> @NonNull ArgumentDescription<C> empty() {
+        return Description.empty();
     }
 
     /**
      * Create a command description instance.
      *
+     * @param <C> Command sender type
      * @param string Command description
      * @return Created command description
      */
     @SuppressWarnings("deprecation")
-    static @NonNull ArgumentDescription of(final @NonNull String string) {
+    static <C> @NonNull ArgumentDescription<C> of(final @NonNull String string) {
         if (requireNonNull(string, "string").isEmpty()) {
-            return Description.EMPTY;
+            return Description.empty();
         } else {
-            return new Description(string);
+            return new Description<>(string);
         }
     }
 
     /**
      * Get the plain-text description.
      *
+     * @param sender Command sender
      * @return Command description
      */
-    @NonNull String getDescription();
+    @NonNull String getDescription(C sender);
 
     /**
      * Get whether or not this description contains contents.
      *
+     * @param sender Command sender
      * @return if this description is empty or not
      */
-    default boolean isEmpty() {
-        return this.getDescription().isEmpty();
+    default boolean isEmpty(final C sender) {
+        return this.getDescription(sender).isEmpty();
     }
 
 }
--- a/cloud-core/src/main/java/cloud/commandframework/Command.java
+++ b/cloud-core/src/main/java/cloud/commandframework/Command.java
@@ -159,7 +159,7 @@ public class Command<C> {
      */
     @Deprecated
     public Command(
-            final @NonNull Map<@NonNull CommandArgument<C, ?>, @NonNull Description> commandArguments,
+            final @NonNull Map<@NonNull CommandArgument<C, ?>, @NonNull Description<C>> commandArguments,
             final @NonNull CommandExecutionHandler<@NonNull C> commandExecutionHandler,
             final @Nullable Class<? extends C> senderType,
             final @NonNull CommandPermission commandPermission,
@@ -180,7 +180,7 @@ public class Command<C> {
      */
     @Deprecated
     public Command(
-            final @NonNull Map<@NonNull CommandArgument<C, ?>, @NonNull Description> commandArguments,
+            final @NonNull Map<@NonNull CommandArgument<C, ?>, @NonNull Description<C>> commandArguments,
             final @NonNull CommandExecutionHandler<@NonNull C> commandExecutionHandler,
             final @Nullable Class<? extends C> senderType,
             final @NonNull CommandMeta commandMeta
@@ -200,7 +200,7 @@ public class Command<C> {
      */
     @Deprecated
     public Command(
-            final @NonNull Map<@NonNull CommandArgument<C, ?>, @NonNull Description> commandArguments,
+            final @NonNull Map<@NonNull CommandArgument<C, ?>, @NonNull Description<C>> commandArguments,
             final @NonNull CommandExecutionHandler<@NonNull C> commandExecutionHandler,
             final @NonNull CommandPermission commandPermission,
             final @NonNull CommandMeta commandMeta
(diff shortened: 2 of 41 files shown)


cloud-core/src/main/java/cloud/commandframework/arguments/flags/CommandFlag.java
```diff
@@ -94,15 +94,17 @@ private CommandFlag(
...
      * <p>
      *
+     * @param sender Command sender
      * @return Flag description
      * @deprecated for removal since 1.4.0. Use {@link #getArgumentDescription()} instead.
      */
     @Deprecated
-    public cloud.commandframework.@NonNull Description getDescription() {
+    @SuppressWarnings("unchecked")
+    public cloud.commandframework.@NonNull Description<?> getDescription(final Object sender) {
```
zml2008, line 103: we can't break api -- also you lose the sender type here

cloud-minecraft/cloud-minecraft-extras/src/main/java/cloud/commandframework/minecraft/extras/RichDescription.java
```diff
@@ -52,56 +51,60 @@
     /**
      * Get an empty description.
      *
+     * @param <C> Command sender type
      * @return the empty description
      */
-    public static @NonNull RichDescription empty() {
-        return EMPTY;
+    public static <C> @NonNull RichDescription<C> empty() {
```
zml2008, line 57: keep the constant, just cast it

cloud-core/src/main/java/cloud/commandframework/ArgumentDescription.java
```diff
@@ -32,49 +32,54 @@
...
     }
 
     /**
      * Get the plain-text description.
      *
+     * @param sender Command sender
      * @return Command description
      */
-    @NonNull String getDescription();
+    @NonNull String getDescription(C sender);
```
zml2008, line 73: you can't break API....

cloud-core/src/main/java/cloud/commandframework/Command.java
```diff
@@ -159,7 +159,7 @@ public Command(
      */
     @Deprecated
     public Command(
-            final @NonNull Map<@NonNull CommandArgument<C, ?>, @NonNull Description> commandArguments,
+            final @NonNull Map<@NonNull CommandArgument<C, ?>, @NonNull Description<C>> commandArguments,
```
zml2008, line 162: every type variable for these should be `<? super C>` -- that makes things more flexible

cloud-core/src/main/java/cloud/commandframework/Command.java
```diff
@@ -358,17 +358,18 @@ public Command(
...
+     * @param sender Command sender
      * @param argument Argument
      * @return Argument description
      * @throws IllegalArgumentException If the command argument does not exist
      * @deprecated More than one matching command argument may exist per command.
      *             Use {@link #getArguments()} and search in that, instead.
      */
     @Deprecated
-    public @NonNull String getArgumentDescription(final @NonNull CommandArgument<C, ?> argument) {
+    public @NonNull String getArgumentDescription(final C sender, final @NonNull CommandArgument<C, ?> argument) {
```
zml2008, line 369: api break

cloud-core/src/main/java/cloud/commandframework/CommandComponent.java
```diff
@@ -65,15 +65,16 @@ private CommandComponent(
     /**
      * Gets the command component description
      *
+     * @param sender Command sender
      * @return command component description
      * @deprecated for removal since 1.4.0. Use {@link #getArgumentDescription()} instead.
      */
     @Deprecated
-    public @NonNull Description getDescription() {
+    public @NonNull Description<C> getDescription(final C sender) {
```
zml2008, line 73: api break

cloud-core/src/main/java/cloud/commandframework/Description.java
```diff
@@ -48,33 +44,36 @@
...
      *
+     * @param <C> Command sender type
      * @return Command description
      * @deprecated for removal since 1.4.0. See {@link ArgumentDescription#empty()}
      */
     @Deprecated
-    public static @NonNull Description empty() {
-        return EMPTY;
+    public static <C> @NonNull Description<C> empty() {
+        return new Description<>("");
```
zml2008, line 53: keep the constant here, just cast it

cloud-core/src/test/java/cloud/commandframework/CommandTreeTest.java
```diff
@@ -264,10 +264,10 @@ void getArgumentsAndComponents() {
...
 
         // Check description is set for all components, is empty when not specified
-        Assertions.assertEquals("", components.get(0).getArgumentDescription().getDescription());
-        Assertions.assertEquals("", components.get(1).getArgumentDescription().getDescription());
-        Assertions.assertEquals("detaildescription", components.get(2).getArgumentDescription().getDescription());
-        Assertions.assertEquals("argumentdescription", components.get(3).getArgumentDescription().getDescription());
+        Assertions.assertEquals("", components.get(0).getArgumentDescription().getDescription(new TestCommandSender()));
+        Assertions.assertEquals("", components.get(1).getArgumentDescription().getDescription(new TestCommandSender()));
+        Assertions.assertEquals("detaildescription", components.get(2).getArgumentDescription().getDescription(new TestCommandSender()));
+        Assertions.assertEquals("argumentdescription", components.get(3).getArgumentDescription().getDescription(new TestCommandSender()));
```
zml2008, line 270: don't re-create new sender instances, re-use one instead

cloud-minecraft/cloud-velocity/src/main/java/cloud/commandframework/velocity/arguments/PlayerArgument.java
```diff
@@ -57,7 +57,7 @@ private PlayerArgument(
             final boolean required,
             final @NonNull String name,
             final @Nullable BiFunction<CommandContext<C>, String, List<String>> suggestionsProvider,
-            final @NonNull ArgumentDescription defaultDescription,
+            final @NonNull ArgumentDescription<C> defaultDescription,
```
zml2008, line 60: `<? super C>`, as with other places

zml2008 (changes requested):
this cannot be merged as-is -- resolve the API breaks (probably add the new sender-context methods as default implementations that delegate to the non-sender-specific ones)

I'm also not sure about your handling in CommandFlag -- we might have to add a sender type parameter to that as well.

not sure about the type parameters -- while at runtime the change won't be breaking, at compile-time we'll have to ensure that the compiler's type inference can resolve the majority of simple cases. While Java supports graceful migration to use generics, Kotlin does not -- raw types are a compile error on that side, so we also have to determine if we consider that enough of a breaking change that this would have to be held for Cloud 2.

cloud-core/src/main/java/cloud/commandframework/Command.java
egg82: I was following the existing patterns with `CommandArgument<C, ?>` here. Would it be preferred to change the method signature to `final @NonNull Map<@NonNull CommandArgument<? super C, ?>, @NonNull Description<? super C>>` ?

cloud-core/src/main/java/cloud/commandframework/arguments/flags/CommandFlag.java
egg82: The entire class isn't generic and uses `?` in place of sender types. Would you prefer the class was generified?

cloud-core/src/main/java/cloud/commandframework/Command.java
egg82: How would you want to handle this case? If the sender is not provided, then would you call the new method with sender with a null value, or would you make up a new sender object that does nothing?

PR #239, opened 2021-03-18
`@Quoted` annotation for string arguments in annotated methods

jpenilla:


jpenilla:
> does it make sense to throw an exception here if both quoted and greedy are applied to one value?

probably yes

zml2008 (approved):

broccolai (approved):

PR #249, opened 2021-04-20
Rename Fabric references to Quilt

broccolai:
I left some of the Fabric references in the documentation whilst we are relying on Fabric-API


kashike (approved):

JarFiles (approved):

Romvnly-Gaming:
I don't understand. If the Fabric implementation relies on Fabric, why are you renaming it?

PR #257, opened 2021-04-29
bukkit: Implement ItemStackArgument

jpenilla:
Implements ItemStackArgument on Bukkit by wrapping the Brigadier ArgumentType on 1.13+, and falling back to the simple Material parser for older versions.

This is a cleaner approach for modern versions than #228 took, however is less functional on legacy versions.

Also in this PR: a little bit of general cleanup to the Bukkit impl


broccolai (approved):
LGTM, @Apistatus.Internal would be nice on CraftBukkitReflection but have two nullability imports would be kinda annoying.

PR #258, opened 2021-04-30
bukkit: Implement BlockDataArgument

jpenilla:
Implements a parser for `BlockData`. This wraps Minecraft's `BlockState` `ArgumentType`.

The Brigadier parser we are wrapping can technically parse NBT data in addition to type and properties, but there isn't exactly a good way to expose a raw `CompoundTag` to the Bukkit API world.

In vanilla, the parsed NBT is only ever used when setting blocks in the world. The return type for the vanilla argument is `BlockInput`, and it has a method which sets a block in the world to the parsed `BlockState`, and then if it is a BlockEntity applies the parsed NBT.

We could probably return a wrapper around `BlockInput`, similar to `ProtoItemStack`, and expose the aforementioned method, however I'm not sure if this is something that would even be desired, or if the wrapper would just be annoying to unbox every time.


broccolai (changes requested):
Fix the checkstyle complaints, LGTM otherwise

PR #272, opened 2021-07-07
Add support for suspending functions

Citymonstret:
This also changes CommandExecutionHandler to return futures instead. The old method is still supported, and the new future-returning method will delegate to the old non-returning method.


NickAcPT (commented):
Took cloud long enough to have this.
LGTM

broccolai:
It's already merged what do you mean :trollface:

PR #279, opened 2021-07-23
Implement flag permissions

FrankHeijden:
This will allow developers to create flags which require a permission to use.


Citymonstret:
Sorry for taking so long to look at this, could you throw `@since` into the JavaDoc of the new functions and possibly add the change to https://github.com/Incendo/cloud/blob/1.6.0-dev/CHANGELOG.md ? 🥺

FrankHeijden:
> Sorry for taking so long to look at this, could you throw `@since` into the JavaDoc of the new functions and possibly add the change to https://github.com/Incendo/cloud/blob/1.6.0-dev/CHANGELOG.md ? 🥺

pin2jxhgz2e51

PR #289, opened 2021-08-18
Correct velocity example not appending error component

a-tort:
Fixes the error message not being appended in the velocity example.


jpenilla (approved):
Thanks!

PR #291, opened 2021-08-28
[JDA] Use Message#getContentDisplay

GriffinCodes:
`Message#getContentRaw` will not resolve any mentionables, such as users and channels. This makes the content `<#241774576822910976> <@115552359458799616>` instead of `#general @Griffin`.

I personally can't think of a reason why you would need the raw format. If you want the mentions, you should use arguments.

Closes https://github.com/ProjectEdenGG/Jayce/issues/1

### DIFF
--- a/cloud-discord/cloud-jda/src/main/java/cloud/commandframework/jda/JDACommandListener.java
+++ b/cloud-discord/cloud-jda/src/main/java/cloud/commandframework/jda/JDACommandListener.java
@@ -69,7 +69,7 @@ public class JDACommandListener<C> extends ListenerAdapter {
         }
 
         final String prefix = this.commandManager.getPrefixMapper().apply(sender);
-        String content = message.getContentRaw();
+        String content = message.getContentDisplay();
 
         if (!content.startsWith(prefix)) {
             return;


p5nbTgip0r:
Wouldn't this break argument parsing for Discord objects? Last I checked the argument parsing for things like `Channel` and `Role` relies on using the raw input to see the object id that was mentioned, but if you use `getContentDisplay` then these ids aren't available anymore.. I could be mistaken though

GriffinCodes:
Hmm, you're right. I'll come up with a different solution.

broccolai:
Perhaps this could be added to the context via the processor?
see: [this](https://github.com/Incendo/cloud/blob/f14d27314beabc944820eb3e42030e56fb01cb54/cloud-discord/cloud-jda/src/main/java/cloud/commandframework/jda/JDACommandPreprocessor.java#L77-L78)

Although the event is already in the context there so theoretically you could already get the content in that format, not sure if want to bloat the context even more but I'm open to it.

GriffinCodes:
I'm not sure how exactly you would implement it in the context.

The only solution that I've come up with that I think sounds "correct" is to replace the content with the display content if the object it is parsing does not implement `IMentionable`, but I'm not familiar enough with the framework to know how to do that, and the tokenized queue complicates that

PR #296, opened 2021-09-13
core: fix And/OrPermission losing conditional tree

Machine-Maker:
If you tried to construct an Or permission from two and permissions, you would lose the `and` part of the conditional tree and just one big or permission would be created.


jpenilla:
Could you add a test case for this to `CommandPermissionTest`?

Machine-Maker:
@jpenilla ok, I added some tests (and fixed the same problem with `AndPermission#of`. You can just copy the tests without the changes and see that they fail.

jpenilla (approved):

Issue #123, opened 2020-10-29
Bukkit uses description from last registered command for a node


YouHaveTrouble:
When commands are displayed in help menu they get a description from the last registered subcommand.
Expected behaviour is to have description from the root module.

Cloud verison: 1.1.0

edit:
This is referring to the Bukkit help menu, which is using the description from the last registered command for a root node, as that command is the one whose Bukkit command instance is registered in the command map for that root node.

PR #297, opened 2021-09-14
feat: delegate bukkit permission check to cloud manager

Machine-Maker:
Its possible this was left out for some reason, and if that's the case well whatever, but I noticed that permission changes were not being reflected in the bukkit help menu, and after some debugging, realized the permission check in BukkitCommand was always true because the permission field was null.


jpenilla:
The issue with this is it will just check the last registered cloud command for that root node (similar issue to #123). Implemented a slightly different fix in https://github.com/Incendo/cloud/commit/fd7b5265eb794d4512012dc98482c93408e3d704

PR #301, opened 2021-09-19
Make CommandMethod applicable to class

FrankHeijden:
Isse nice for not having to repeat the base command :)

### DIFF
--- a/cloud-annotations/src/main/java/cloud/commandframework/annotations/AnnotationParser.java
+++ b/cloud-annotations/src/main/java/cloud/commandframework/annotations/AnnotationParser.java
@@ -406,7 +406,8 @@ public final class AnnotationParser<C> {
             final @NonNull Object instance,
             final @NonNull Collection<@NonNull CommandMethodPair> methodPairs
     ) {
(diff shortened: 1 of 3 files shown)


cloud-annotations/src/main/java/cloud/commandframework/annotations/AnnotationParser.java
```diff
@@ -406,13 +406,16 @@ public void registerCommandExecutionMethodFactory(
             final @NonNull Object instance,
             final @NonNull Collection<@NonNull CommandMethodPair> methodPairs
     ) {
+        final CommandMethod classCommandMethod = instance.getClass().getAnnotation(CommandMethod.class);
```
Citymonstret, line 409: Might seem random, but could you do this:
```java
final AnnotationAccessor classAnnotations = AnntationAccessor.of(instance.getClass());
final CommandMethod classCommandMethod = classAnnotation.get(CommandMethod.class);
```
? 

It'd be nice to be consistent about where we retrieve annotations from

FrankHeijden: Fixed

PR #303, opened 2021-09-27
annotations: Also apply builder modifiers from class

jpenilla:
Current behavior allows applying the same builder modifier to a method multiple times by using the annotation more than once.

I think it might be more intuitive if a builder modifier applied to the class was only used on methods which don't also have that modifier applied (ex below), but I didn't make that change to keep the aforementioned behavior of repeating annotations.

```java
@MyAnnotation("Class default")
class Commands {
  @CommandMethod(...)
  void someCommandUsingClassDefaultForMyAnnotation(...) {}

  @MyAnnotation("Different value")
  @CommandMethod(...)
  void usesSpecificValueInstead(...) {}
}
```

That kind of behavior is still possible as the class builder modifiers will be applied before the method specific ones, but both modifiers will apply.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Changed
 - Added `executeFuture` to `CommandExecutionHandler` which is now used internally. By default, this delegates to the old 
   `execute` method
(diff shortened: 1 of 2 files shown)


Citymonstret:
> That kind of behavior is still possible as the class builder modifiers will be applied before the method specific ones, but both modifiers will apply.

I think this is fine, as the annotation access is ordered and the method annotations would get priority. It's more versatile this way :shrug:

Citymonstret (approved):

PR #304, opened 2021-09-27
Make CommandExecutionHandler#executeFuture return CompletableFuture<Void> instead of CompletableFuture<Object>

jpenilla:
The value of the completed future is never used, void makes this more obvious


solonovamax:
Why was this merged into 1.6.0?

This is a breaking API change, shouldn't that be left for the next major version?

Because, if you have a class that extends it which returned `Object`, it will no longer compile.

solonovamax:
Yeah, I'm just saying that when 1.6.0 is eventually released, it'd be a breaking change, going against the semver spec, no?
Because, you can't introduce any breaking changes between minor versions, so any code from `1.5.0` must compile for `1.6.0`.

jpenilla:
executeFuture did not exist in 1.5

solonovamax:
ah, nevermind then

PR #307, opened 2021-10-01
Fix cloud swallowing exceptions in `suspend` methods.

### LINKED ISSUE #306: [1.6.0] Kotlin `suspend` functions swallow exceptions with Asynchronous execution
Kotlin `suspend` functions will sometimes swallow exceptions when using `AsynchronousCommandExecutionCoordinator`

This happens because of the following:

This all happens in the `AsynchronousCommandExecutionCoordinator` class.

first, on line `117`, the `commandConsumer` is executed.
next, on line `84`, it executes the command future, but then it *doesn't return that future*
it then returns to line `117`, since that was an anonymous lambda constructed earlier, and after that it *immediately completes the future* on line `118`, so when the `whenComplete` on line `84` is later executed, and it attempts to run `completeExceptionally` on the future, it has *already completed the future*

So basically, it completes the future without waiting for the `commandConsumer`.
Which causes it to be completed *before* we know if the completion is successful or not.

So, there are two things that could be done:
1. have the `commandConsumer` throw the exception (or, just don't use `whenComplete` and actually block on the completion of the command)
2. move `resultFuture.complete` into the `commandConsumer` so we can wait for execution to finish.

Will PR with changes in a few few mins.

solonovamax:
Fixes #306.

Moves the `resultFuture.complete` invocation to the `commandConsumer` lambda rather than completing it before the command execution is finished.

Previously, the `resultFuture` was being `complete`d immediately after `commandConsumer` was invoked, and since `commandConsumer` runs in a non-blocking manner, it swallowed the exception.

Therefore, `resultFuture.complete` was moved into the `commandConsumer`, specifically within `whenComplete`. That way, it would only be completed when the execution has been finished.

There was also some other minor refactors, so that there's only one `return resultFuture`, which occurs at the end of the method, just to keep the code less complex.

### DIFF
--- a/cloud-core/src/main/java/cloud/commandframework/execution/AsynchronousCommandExecutionCoordinator.java
+++ b/cloud-core/src/main/java/cloud/commandframework/execution/AsynchronousCommandExecutionCoordinator.java
@@ -51,16 +51,6 @@ public final class AsynchronousCommandExecutionCoordinator<C> extends CommandExe
     private final Executor executor;
     private final boolean synchronizeParsing;
 
-    /**
-     * Create a new {@link Builder} instance
-     *
-     * @param <C> Command sender type
-     * @return Builder
-     */
-    public static <C> @NonNull Builder<C> newBuilder() {
-        return new Builder<>();
-    }
-
     private AsynchronousCommandExecutionCoordinator(
             final @Nullable Executor executor,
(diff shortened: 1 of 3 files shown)


solonovamax:
Updated with request changes
image

solonovamax:
^ Force pushed to squash commits

PR #313, opened 2021-10-13
Support AND & OR operators in CommandPermission annotation

0utplay:
This PR adds support for AND ("&") & OR ("|") operators in CommandPermission annotation.
The code is tested and works as expected, the new tests for the code pass too.


0utplay:
I thought that this would be the most convenient way of integrating this as aliases are done the same way.

PR #314, opened 2021-10-31
Test the sender type when building a brigadier command

### LINKED ISSUE #202: Command senders are suggested commands which they cannot use
If a specific sender type is set for a command, this is not taken into account for suggestions. For example, if the specific sender type is `Player`, our command will still be suggested in the console, even though console cannot execute this command. Command syntax strings have a similar issue with showing commands not usable by that user.

syldium:
Add a sender type check when building a Brigadier command (fixes #202).

Tested on Paper:
gib in red


jpenilla:
This is more of a hack than a proper fix, it doesn't address the root issue and only makes any difference when brigadier is in use. The BrigadierManager is not the correct place for this fix. (it's possible the fix may involve modifying the brig manager, I haven't looked into it enough to be sure, however I am certain code outside the brig manager will need to be touched as well)

jpenilla (changes requested):
see comment

jpenilla:
This will have issues with shared chains as brig requirements are node-scoped. #629 solves this properly in 2.0.

PR #315, opened 2021-11-10
Allow parser parameter annotations for flags as well

derklaro:
Evening,

i just stumbled over this behaviour of cloud to not allow argument annotations like `@Quoted` or `@Greedy` for flags. I have a specific use case for that. A user should be able using our template installer to define an executable for the build step. The following method to handle the command is used:

```java
  @CommandMethod("template|t install <template> <versionType> <version>")
  public void installTemplate(
    CommandSource source,
    @Argument("template") ServiceTemplate serviceTemplate,
    @Argument("versionType") ServiceVersionType versionType,
    @Argument("version") ServiceVersion serviceVersion,
    @Flag("force") boolean forceInstall,
    @Flag("executable") @Quoted String executable
  ) {
  ```
  
  The executable should be the full path to a java installation (in my case on windows `java.exe`). When I now tried to run the following command:
 `t install local:Lobby/default magma 1.12.2 --executable "C:\Program Files\Java\jdk1.8.0_241\bin\java.exe"`
 
 cloud failed with:
 `FlagArgument$FlagParseException: No flag started. Don't know what to do with 'Files\Java\jdk1.8.0_241\bin\java.exe"'`
 
I feel like all custom annotation parsers should also be available for flags, especially because the types are not changing from when they are an argument but are "more" optional... I hope this explaination was understandable :)

Have a nice evening!

### DIFF
--- a/cloud-annotations/src/main/java/cloud/commandframework/annotations/FlagExtractor.java
+++ b/cloud-annotations/src/main/java/cloud/commandframework/annotations/FlagExtractor.java
@@ -28,14 +28,15 @@ import cloud.commandframework.CommandManager;
 import cloud.commandframework.arguments.CommandArgument;
 import cloud.commandframework.arguments.flags.CommandFlag;
 import cloud.commandframework.arguments.parser.ArgumentParser;
-import cloud.commandframework.arguments.parser.ParserParameters;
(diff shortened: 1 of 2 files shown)


jpenilla (approved):
Looks good, thanks

PR #316, opened 2021-11-14
Split kotlin modules

jpenilla:
Splits the kotlin module into `cloud-kotlin-extensions` and `cloud-kotlin-coroutines-annotations`. The main purpose of this is to not add kotlin-reflect as a dependency to consumers of the existing kotlin module with 1.6.0. I'm open for suggestions of better names than `cloud-kotlin-coroutines-annotations`, but I purposefully left `cloud-kotlin-coroutines` available for a potential non-annotations coroutines module in the future, which wouldn't need kotlin-reflect.


broccolai (approved):
lgtm

PR #332, opened 2022-01-04
fix velocity build

zml2008:


jpenilla:
Could we just bump velocity api to 3.x instead?

zml2008:
likely, but that would also technically be a breaking change, and likely bump `cloud-velocity` to Java 11

jpenilla:
It looks like velocity-api is currently compileOnly, so just bumping the version shouldn't make any difference for consumers. As far as the java version, we should just be able to disableAutoTargetJvm()? Excluding these dependencies works, but if we can I'd prefer to  update to a supported version of velocity-api instead

jpenilla (approved):

PR #335, opened 2022-01-12
Add `PotionEffectTypeArgument`

TehBrian:
Used `EnchantmentArgument` and `DurationArgument` as inspiration, and also included the updates from [jmp's DurationArgument cleanup](https://github.com/Incendo/cloud/commit/f029e846af988294eb1d3050b3cea0e658db9485). 

Works as expected on Paper 1.18.1, please let me know if more testing on different versions should be done.


TehBrian:
how would that be done since the api version that cloud is using doesn't provide PotionEffectType.getByKey?

jpenilla:
You would bump the API to a version which provides that method, and guard usage of it with a check for the existence of the API using reflection. Something similar is done in the KeyedWorldArgument.

TehBrian:
unless anyone else wants this go through, I'd be fine with letting this pr die. the project that was going to use this feature is now using an enum wrapper instead, because bukkit's PotionEffectType naming sucks and doesn't align with modern mc.

TehBrian:
closing for now, feel free to reopen later if anyone wants.

PR #336, opened 2022-01-18
make BooleanArgument suggestions lowercase

TehBrian:


cloud-core/src/main/java/cloud/commandframework/arguments/standard/BooleanArgument.java
```diff
@@ -154,10 +155,16 @@ private Builder(final @NonNull String name) {
...
 
+        private static final List<String> STRICT = Arrays.asList("TRUE", "FALSE");
         private static final List<String> LIBERAL = Arrays.asList("TRUE", "YES", "ON", "FALSE", "NO", "OFF");
         private static final List<String> LIBERAL_TRUE = Arrays.asList("TRUE", "YES", "ON");
         private static final List<String> LIBERAL_FALSE = Arrays.asList("FALSE", "NO", "OFF");
 
+        private static final List<String> STRICT_LOWER = STRICT
+                .stream().map(String::toLowerCase).collect(Collectors.toList());
+        private static final List<String> LIBERAL_LOWER = LIBERAL
+                .stream().map(String::toLowerCase).collect(Collectors.toList());
```
portlek, line 166: please use these kind of methods with Locale.ROOT not raw. this causes weird bugs in Turkish languaged PCs. `toLowerCase(Locale.ROOT)`

portlek (changes requested):

cloud-core/src/main/java/cloud/commandframework/arguments/standard/BooleanArgument.java
TehBrian: good catch, thank you!

Citymonstret (approved):

PR #338, opened 2022-02-07
Update commodore version

lucko:
Works around an issue in Forge which prevents the client from running commands without the executable flag set. (https://github.com/lucko/commodore/commit/7bbcbe4f4722d5b670a074a0995bf24633d03885)

https://github.com/MinecraftForge/MinecraftForge/issues/8438

### DIFF
--- a/README.md
+++ b/README.md
@@ -211,7 +211,7 @@ Here are some projects that are using cloud:
 - [Incendo/bot](https://github.com/Incendo/bot): IRC bot for #incendo on esper.net
 - [Incendo/ban](https://github.com/Incendo/ban): Velocity punishment suite
 - [broccolai/tickets](https://github.com/broccolai/tickets): Minecraft ticket system
-- [pl3xgaming/Pl3xMap](https://github.com/pl3xgaming/Pl3xMap): Minimalistic and lightweight world map viewer for Paper servers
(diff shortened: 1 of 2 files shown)


jpenilla:
I don't think this actually makes any difference with cloud's usage of commodore: https://github.com/Incendo/cloud/blob/master/cloud-minecraft/cloud-bukkit/src/main/java/cloud/commandframework/bukkit/CloudCommodoreManager.java
The `CloudBrigadierManager` should be creating a proper tree where nodes which are executable are set as such, if not that would be an issue with cloud that needs to be fixed, as our brigadier code is reused across other platforms not using commodore.

I have no problem with merging this regardless if it's updated to target the correct branch (1.7.0-dev)

lucko:
Ah yep you're right, sorry I didn't look too closely at your usage.

This bugfix is the only change in the 1.12 release, so probably no point in updating - but I'll update the target branch and leave you to decide that.

AlessioGr:
Uh I don't know if this is related at all, but a user of my plugin previously reported that a certain subcommand which was not registered with cloud, but "catched"/handled with the PlayerCommandPreProcessEvent, would throw a brigardier error Instead of being handled by that event:
unknown-14

The issue was gone once he uninstalled forge and his mods and moved back to vanilla. (The command executed fine on vanilla)

Could this be related at all?

jpenilla:
manual merged to 1.7.0-dev

PR #331, opened 2021-12-21
fix typos, gradle wrapper 7.3.1 -> 7.4

TehBrian:

PR #344, opened 2022-02-23
docs: fix typos

TehBrian:
re-pr of #331. for some reason the old pr wasn't updating - the branch didn't seem linked correctly.

### DIFF
--- a/cloud-core/src/main/java/cloud/commandframework/Command.java
+++ b/cloud-core/src/main/java/cloud/commandframework/Command.java
@@ -1112,9 +1112,9 @@ public class Command<C> {
 
         /**
          * Make the current command be a proxy of the supplied command. This means that
-         * all of the proxied commands variable command arguments will be inserted into this
+         * all of the proxied command's variable command arguments will be inserted into this
          * builder instance, in the order they are declared in the proxied command. Furthermore,
-         * the proxied commands command handler will be showed by the command that is currently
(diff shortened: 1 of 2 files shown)


zml2008:
Looks ilke you're going to have to rebase this -- probably best to target 1.7.0-dev branch as well.

broccolai:
Changing the version in the gradle.properties is not the correct way to update Gradle.
You can run `./gradlew wrapper --gradle-version=<version>`, the update the wrapper.

Also in the future, unless you have a reason for needing the update (like a new java version not working with our wrapper), don't worry about update the wrapper, core team will do it whenever we think it's necessary

TehBrian:
I did run `./gradlew wrapper --gradle-version=7.4`

jpenilla:
it's sometimes needed to run the task twice to ensure the wrapper jar is up to date - the jar doesn't always change with every gradle update - but I know it did with 7.4 and this pr doesn't touch the wrapper jar.

for the reason broccolai mentioned though the gradle update should be dropped from this pr.

on a related note we prefer if prs are kept to address a single thing rather than grouping two unrelated sets of changes

TehBrian:
odd, I just ran both `./gradlew wrapper --gradle-version=7.4` and `./gradlew wrapper` several times and there's no difference. anyway, I'll drop the wrapper commit.

Citymonstret (approved):

PR #355, opened 2022-05-23
build: switch to spotless

zml2008:
since licenser is dead

this takes over running ktlint as well, and adds a few other simple checks


alexandersoderberg-paf (approved):
I didn't know you could handle license headers using Spotless. That's pretty neat :D

Citymonstret (approved):
Double approval time. Woo

PR #356, opened 2022-05-23
fabric: Support 1.19

zml2008:
This drops support for older Minecraft versions.

The CommandRegistrationContext changes on the MC side are a bit awkward as well -- I tried to integrate them into Cloud's parsers without breaking any existing api, but the way the context is collected is quite fragile. It currently doesn't work to pass registry access along to client commands either.

@jpenilla any ideas on how to pass this along to various arguments better?


zml2008:
@jpenilla you probably need per-CommandManager keying on the build context supplier - otherwise you might run into issues with multiple mods using cloud at once

PR #369, opened 2022-06-07
feat(core): support root command deletion & standardize capabilities

Citymonstret:


cloud-core/src/main/java/cloud/commandframework/internal/CommandRegistrationHandler.java
```diff
@@ -54,6 +55,16 @@ public interface CommandRegistrationHandler {
      */
     boolean registerCommand(@NonNull Command<?> command);
 
+    /**
+     * Requests that the given {@code rootCommand} should be unregistered.
+     *
+     * @param rootCommand The command to delete
+     * @since 1.7.0
+     */
+    default void unregisterRootCommand(final @NonNull StaticArgument<?> rootCommand) {
```
zml2008, line 64: Should this default to a no-op instead? the feature is already guarded by a capability for managers.

zml2008 (changes requested):

PR #375, opened 2022-06-09
Enable sender aware argument descriptions

Phinner:
This should do the trick for #237, but revapi is still screaming at me.


Phinner:
Discord_9oqZLtjyuU

Yes, it's maybe better to leave it for now...

PR #376, opened 2022-06-10
bukkit/paper: Add NamespacedKeyArgument

jpenilla:


cloud-minecraft/cloud-bukkit/src/main/java/cloud/commandframework/bukkit/argument/NamespacedKeyArgument.java
```diff
@@ -0,0 +1,350 @@
...
+
+    }
+
+
+    /**
+     * Exception used when {@link Parser} fails.
+     *
+     * @since 1.7.0
+     */
+    public static final class NamespacedKeyParseException extends ParserException {
```
Citymonstret, line 297: This is a note, more so than a request. In `cloud-core` I've started generating `equals`, `hashCode` and `toString` for the various parser exceptions. This means that we can then use stuff like
```java
        assertThat(result.getFailure()).hasValue(new NamespacedKeyArgument.NamespacedKeyParseException(
                caption,
                liberal,
                this.context
        ));
```
`toString()` in this case means that truth can differentiate between identical string-representations even when `equals` fails, which gives a slightly better failure mesage.

However, because we haven't really unit tested `cloud-bukkit`, it's less important here.

cloud-minecraft/cloud-bukkit/src/main/java/cloud/commandframework/bukkit/BukkitParserParameters.java
```diff
@@ -0,0 +1,65 @@
...
+ *
+ * @since 1.7.0
+ */
+public final class BukkitParserParameters {
+
+    private BukkitParserParameters() {
+    }
+
+    /**
+     * Sets to require explicit namespaces for {@link cloud.commandframework.bukkit.argument.NamespacedKeyArgument}
```
Citymonstret, line 41: Sets -> Set?

cloud-minecraft/cloud-bukkit/src/main/java/cloud/commandframework/bukkit/BukkitParserParameters.java
```diff
@@ -0,0 +1,65 @@
...
+     * (i.e. 'test' will be rejected but 'test:test' will pass).
+     *
+     * @since 1.7.0
+     */
+    public static final ParserParameter<Boolean> REQUIRE_EXPLICIT_NAMESPACE =
+            create("require_explicit_namespace", TypeToken.get(Boolean.class));
+
+    /**
+     * Sets ta custom default namespace for {@link cloud.commandframework.bukkit.argument.NamespacedKeyArgument}.
```
Citymonstret, line 50:
```suggestion
     * Sets a custom default namespace for {@link cloud.commandframework.bukkit.argument.NamespacedKeyArgument}.
```

Citymonstret (commented):
lgtm :) a very small typo tho

PR #377, opened 2022-06-13
chore(core): deprecate prefixed accessors/mutators in CommandManager

Citymonstret:
All prefixed (actual) getters/setters in CommandManager have been deprecated, and non-prefixed alternatives have been introduced. I've also put some effort into improving the JavaDocs of these methods.


Phinner:
Why doing that ?

Citymonstret:
> Why doing that ?

@Phinner Because right now we have a mixture of both prefixed, and non-prefixed methods. We should settle for one format, and this is the one that Java has decided to use, so we'll follow lead.

Phinner:
Oh really ? Ok, I just though it the inverse.

PR #380, opened 2022-06-20
expose failure reason when flag parsing fails

derklaro:
Makes it easier to find the issue when flag parsing fails, especially when using custom exception handling.


Citymonstret (approved):

cloud-core/src/main/java/cloud/commandframework/arguments/compound/FlagArgument.java
```diff
@@ -514,6 +516,17 @@ public FlagParseException(
         public String getInput() {
             return this.input;
         }
+
+        /**
+         * Get the reason why the flag parsing failed
```
Citymonstret, line 521:
```suggestion
         * Returns the reason why the flag parsing failed.
```

PR #381, opened 2022-06-29
Expose argument names to minecraft extras help message provider

jpenilla:
As noted this only properly works for arguments used a single time (as backwards compat uses arg index), however this works for all current arguments.


broccolai (approved):
LGTM, needs to be rebased and changed to 1.9 though

jpenilla:
Done in 2.0

PR #382, opened 2022-06-29
Don't show commands the sender can't use in invalid syntax messages

jpenilla:
Needs clean up but this is something we should fix.
Should we also check sender type possibly?


jpenilla:
Somehow forgot this existed when creating #416, not certain which I like more

Machine-Maker:
> not certain which I like more

I think I prefer the other solution to this. All the casts to SenderAware in CommandTree are annoying.

PR #389, opened 2022-08-23
fix root command unregister

derklaro:
The command unregistration process unregisters all command nodes from the backing tree based on all childs of a root command. As the Node#equals & Node#hashCode functions are only implemented based on the name of a node, unregistering a command which has (for example) the same literal argument name as a root node will cause the unregister process to remove the root node instead of the child from the parent node.

The root boolean which is now passed to the `deleteRecursively` method seems to be the safest way to check if the specified node should be unregistered from the root command tree, however checking if `node.getParent == internalTree` would do the job as well.


Citymonstret (approved):
LGTM. Thanks for your contribution :)

Issue #230, opened 2021-02-23
Allow flags to be used in the middle of commands.


MCMDEV:
I am aware of whats written in #75 but I think its incredibly annoying. Maybe it's possible to allow this by excluding brigadier from parsing flags or disable brigadier alltogether by an option.

MCMDEV:
I've also just noticed that WorldEdit's Fabric and Forge implementation allows flags anywhere and uses brigadier

Citymonstret:
If anyone can figure out a good solution to this that doesn't inflate the command tree into insane sizes, then I'd be more than willing to add this to the library.

The only solution I've been able to think of so far includes attaching all permutations of the flags, something along the lines of:
```
/foo a b c —d —e 
/foo a —d b c —e
/foo a —d b —e c
/foo a —d —e —b c
/foo a b c —e —d
/foo a —e b c —e
/foo a —e b —d c
/foo a —e —d b c
```

This will be _fine_, as long as people don't have an insane amount of flags in the tree.

Pablete1234:
Would making the "permutations" be `/foo [flags] a [flags] b [flags] c [flags]` feasible? simply an optional node in between each param that tries to parse flags, the flags parser being the one in charge for checking for as many flags as there are there?

Pablete1234:
I've put together a proof-of-concept version that seems to pass all tests doing basically what i described above.

 - The command when being added to the tree inserts the flags argument between every argument
 - The flag parser was changed to support not parsing anything (non-greedy flag parsing, only parsing if the text in the queue starts with "-"). 
  - A small workarround is required for tab suggestions (ie: `/foo` has to suggest `a`, but normally only [flags] would be suggested) to add the next param suggestions if it's a flag.

https://github.com/Incendo/cloud/commit/430beb7e759c72b8949db276fe985f44f1c72fea

It is very much just a proof of concept, it does a few dirty things, and i believe it will have some bugs (probably duplicate flags aren't detected since the HashSet is created each time parse is called, but should be easy to workarround if state is fully obtained from command context)

Issue #190, opened 2020-12-21
Greedy parsers cannot have functioning suggestions


bergerkiller:
There's a hard limitation that a parser can only provide suggestions for one argument being typed. This is because as input to suggestions, only a single String can be provided. When there are two or more parts in the command queue, it is unclear what part should be sent to the suggestions provider.

A greedy string parser might want to suggest words with spaces in them, while a string[] parser might want to suggest a word for every element.

Because suggestions aren't even asked right now (because it is not clear what should happen), there is no workable solution.

https://github.com/Incendo/cloud/blob/1.4.0-dev/cloud-core/src/main/java/cloud/commandframework/CommandTree.java#L602

CompoundArgument has some special logic tailored to it, but also assumes the last-typed 'element' is what needs suggestions. For greedy parsers that is not valid.

Changing String to Queue\<String\> for suggestions would make it in parity with parsers, but would also break api compatibility. I know no good solution for this.

Pablete1234:
### Non-breaking solutions

#### Option A (small behavior change for greedy parsers)
Make it so a greedy or flag-yielding-greedy parser, will receive a string that is concatenating the remaining input with spaces. Non-greedy parsers will not be affected, and the greedy parsers that do get affected were already broken or mostly non-functioning, so this could be an ok solution.

#### Option B (no behavior change)
The command context could have a property for the currently-parsing-queue, similar to how getRawInput is a thing, would allow the flexibility of using the remaining text for greedy-parser suggestions without compromising breaking behavior anywhere.


### Partial workarround:
A way to work around this issue from within the `suggestions` method is to use the command context's raw input to re-create the real string.
Example (call with commandContext.getRawInput(), input):
```java
  private List<String> getRemainingInput(LinkedList<String> rawInput, String input) {
    if (!options.has(StandardParameters.GREEDY) && !options.has(StandardParameters.FLAG_YIELDING))
      return Collections.singletonList(input);

    List<String> value = new ArrayList<>();

    Iterator<String> it = rawInput.descendingIterator();
    while (it.hasNext()) {
      String val = it.next();
      value.add(val);
      if (val.equals(input)) break;
    }
    Collections.reverse(value);
    return value;
  }
```
[...]

PR #395, opened 2022-10-16
Add option to allow flags anywhere after last literal argument

Pablete1234:
This is a proof-of-concept of how the library could support flags being located anywhere in the command, instead of just at the very end.

The code at this time is dirty just but the main point is to open up discussion about do's and don'ts since i'm new to the library, tests are passing (including the newly added ones) and the changes seem to be working as intended, although i'm sure there's missing corners).

Solves #230

### DIFF
--- a/cloud-core/src/main/java/cloud/commandframework/Command.java
+++ b/cloud-core/src/main/java/cloud/commandframework/Command.java
@@ -62,6 +62,7 @@ public class Command<C> {
 
     private final List<@NonNull CommandComponent<C>> components;
     private final List<@NonNull CommandArgument<C, ?>> arguments;
+    private final @Nullable FlagArgument<C> flagArgument;
     private final CommandExecutionHandler<C> commandExecutionHandler;
     private final Class<? extends C> senderType;
     private final CommandPermission commandPermission;
@@ -78,6 +79,7 @@ public class Command<C> {
      * @since 1.3.0
      */
     @API(status = API.Status.STABLE, since = "1.3.0")
+    @SuppressWarnings("unchecked")
     public Command(
             final @NonNull List<@NonNull CommandComponent<C>> commandComponents,
             final @NonNull CommandExecutionHandler<@NonNull C> commandExecutionHandler,
@@ -90,6 +92,14 @@ public class Command<C> {
         if (this.components.isEmpty()) {
             throw new IllegalArgumentException("At least one command component is required");
         }
+
+        this.flagArgument =
+                this.arguments.stream()
+                        .filter(ca -> ca instanceof FlagArgument)
+                        .map(ca -> (FlagArgument<C>) ca)
+                        .findFirst()
+                        .orElse(null);
+
         // Enforce ordering of command arguments
         boolean foundOptional = false;
         for (final CommandArgument<C, ?> argument : this.arguments) {
@@ -316,6 +326,32 @@ public class Command<C> {
      */
     public @NonNull List<CommandArgument<@NonNull C, @NonNull ?>> getArguments() {
         return new ArrayList<>(this.arguments);
+    }
+
+    /**
+     * Return a mutable copy of the command arguments, ignoring flag arguments.
+     *
+     * @return argument list
+     * @since 1.8.0
+     */
+    @API(status = API.Status.EXPERIMENTAL, since = "1.8.0")
+    public @NonNull List<CommandArgument<@NonNull C, @NonNull ?>> nonFlagArguments() {
+        List<CommandArgument<C, ?>> arguments = new ArrayList<>(this.arguments);
+        if (this.flagArgument != null) {
+            arguments.remove(this.flagArgument);
+        }
+        return arguments;
+    }
+
+    /**
+     * Returns the flag argument for this command, or null if no flags are supported.
+     *
+     * @return flag argument or null
+     * @since 1.8.0
+     */
+    @API(status = API.Status.EXPERIMENTAL, since = "1.8.0")
+    public @Nullable FlagArgument<@NonNull C> flagArgument() {
+        return this.flagArgument;
     }
 
     /**
--- a/cloud-core/src/main/java/cloud/commandframework/CommandManager.java
+++ b/cloud-core/src/main/java/cloud/commandframework/CommandManager.java
@@ -1420,14 +1420,14 @@ public abstract class CommandManager<C> {
         OVERRIDE_EXISTING_COMMANDS,
 
         /**
-         * Allows parsing flags at any position by appending flag argument nodes between each command node.
+         * Allows parsing flags at any position after the last literal by appending flag argument nodes between each command node.
          * It can have some conflicts when integrating with other command systems like Brigadier,
          * and code inspecting the command tree may need to be adjusted.
          *
          * @since 1.8.0
          */
         @API(status = API.Status.EXPERIMENTAL, since = "1.8.0")
-        ALLOW_FLAGS_EVERYWHERE
+        LIBERAL_FLAG_PARSING
     }
 
 
--- a/cloud-core/src/main/java/cloud/commandframework/CommandManager.java
+++ b/cloud-core/src/main/java/cloud/commandframework/CommandManager.java
@@ -1417,7 +1417,17 @@ public abstract class CommandManager<C> {
          * @since 1.2.0
          */
         @API(status = API.Status.STABLE, since = "1.2.0")
-        OVERRIDE_EXISTING_COMMANDS
+        OVERRIDE_EXISTING_COMMANDS,
+
+        /**
+         * Allows parsing flags at any position by appending flag argument nodes between each command node.
+         * It can have some conflicts when integrating with other command systems like Brigadier,
+         * and code inspecting the command tree may need to be adjusted.
+         *
+         * @since 1.8.0
+         */
+        @API(status = API.Status.EXPERIMENTAL, since = "1.8.0")
+        ALLOW_FLAGS_EVERYWHERE
     }
 
 
--- a/cloud-core/src/main/java/cloud/commandframework/CommandTree.java
+++ b/cloud-core/src/main/java/cloud/commandframework/CommandTree.java
@@ -339,7 +339,8 @@ public final class CommandTree<C> {
                 ));
             }
             if (child.getValue() != null) {
-                if (commandQueue.isEmpty()) {
+                // Flag arguments need to be skipped over, so that further defaults are handled
+                if (commandQueue.isEmpty() && !(child.getValue() instanceof FlagArgument)) {
                     if (child.getValue().hasDefaultValue()) {
                         commandQueue.add(child.getValue().getDefaultValue());
                     } else if (!child.getValue().isRequired()) {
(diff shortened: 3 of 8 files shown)


Pablete1234:
I've cleaned-up the code to a state where it's stable, i'd love to get some reviews or comments regarding it.

cloud-core/src/main/java/cloud/commandframework/Command.java
```diff
@@ -318,6 +328,28 @@ public Command(
         return new ArrayList<>(this.arguments);
     }
 
+    /**
+     * Return a copy of the command argument array, ignoring flag arguments.
+     *
+     * @return Copy of the command argument array, ignoring flag arguments.  This List is mutable.
+     */
+    public @NonNull List<CommandArgument<@NonNull C, @NonNull ?>> getActualArguments() {
```
jpenilla, line 336: I see two options here
- 1
   - make getArguments return a legacy view of the arguments (flags at end only)
   - add rawArguments to get the raw arguments with potential repeated flags
- 2
   - check all internal usages (ie help) and make it properly handle the new actualArguments and flagArgument instead of relying on getArguments
   - expect users understand they may need to change other code when enabling flags anywhere
 
I'm not sure if other places like syntax messages and help menus need updating for "correct" behavior when the tree has flagnodes throughout (ie none of the aforementioned options address accessing the command tree, only the arguments through the command instance), so that probably also needs to be addressed.

jpenilla (changes requested):

cloud-core/src/main/java/cloud/commandframework/CommandTree.java
```diff
@@ -638,8 +637,7 @@ private CommandTree(final @NonNull CommandManager<C> commandManager) {
             }
             return Collections.emptyList();
         } else if (commandQueue.peek().isEmpty()) {
-            commandContext.setCurrentArgument(child.getValue());
-            return child.getValue().getSuggestionsProvider().apply(commandContext, commandQueue.remove());
+            return this.directSuggestions(commandContext, child, commandQueue.peek());
```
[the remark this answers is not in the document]
Pablete1234: the change is intentional, the removal was inconsistent across the different branches of the if-statement, and it resulted on differing behavior for no good reason. The remaining argument shouldn't be consumed off of the queue because the suggestion parser may want to access it, and although that's more related to #190 it's still a step towards enabling that to work more reliably.

cloud-core/src/main/java/cloud/commandframework/Command.java
Pablete1234: To maintain backwards-compat and not break API, i've made getArguments as well as the arguments inside stay the same, with flags at the end

The new method (getActualArguments) is the list of arguments without the flags, which is just a util to ignore flags. The getArguments method isn't considered a "legacy view" but rather still the "normal" view where flags will be at the end.

The flags being available between all args is only a detail relevant to the command tree registration, but the help generation/syntax will never display the command as having the flags everywhere because it isn't the case.

I believe this solution makes it so most api users can stay ignorant of the implementation detail that parsing will support flags everywhere, help menus etc would look absurdly verbose if they were displayed as such.

cloud-core/src/main/java/cloud/commandframework/CommandManager.java
```diff
@@ -1417,7 +1417,16 @@ public enum ManagerSettings {
...
-        OVERRIDE_EXISTING_COMMANDS
+        OVERRIDE_EXISTING_COMMANDS,
+
+        /**
+         * When enabled, optional flag arguments are appended between each command node.
+         * It can have some conflicts when integrating with other command systems like brigadier.
+         *
+         * @since 1.8.0
+         */
+        @API(status = API.Status.STABLE, since = "1.8.0")
```
jpenilla, line 1428: probably go with experimental here as well if we do on the other related new methods

cloud-core/src/main/java/cloud/commandframework/context/CommandContext.java
```diff
@@ -558,6 +559,24 @@ public void remove(final @NonNull CloudKey<?> key) {
         return this.getOptional(key).orElseGet(defaultSupplier);
     }
 
+    /**
+     * Get a value if it exists, else compute and store the value by the supplier and return it
```
jpenilla, line 563:
```suggestion
     * Get a value if it exists, else compute and store the value returned by the function and return it.
```

cloud-core/src/main/java/cloud/commandframework/context/CommandContext.java
```diff
@@ -558,6 +559,24 @@ public void remove(final @NonNull CloudKey<?> key) {
...
+     *
+     * @param key             Argument key
+     * @param defaultSupplier Supplier of default value
+     * @param <T>             Argument type
+     * @return Argument, or supplied default value
+     * @since 1.8.0
+     */
+    public <T> T computeIfAbsent(
```
jpenilla, line 571: `@API`, this one should be stable

cloud-core/src/main/java/cloud/commandframework/Command.java
```diff
@@ -318,6 +328,32 @@ public Command(
         return new ArrayList<>(this.arguments);
     }
 
+    /**
+     * Return a copy of the command argument array, ignoring flag arguments.
+     *
+     * @return Copy of the command argument array, ignoring flag arguments. This List is mutable.
```
jpenilla, line 334: Repeating the main doc in the return is redundant, something simple like `argument list` should suffice.

cloud-core/src/main/java/cloud/commandframework/Command.java
```diff
@@ -318,6 +328,32 @@ public Command(
...
+        if (this.flagArgument != null) {
+            arguments.remove(this.flagArgument);
+        }
+        return arguments;
+    }
+
+    /**
+     * Returns the flag argument for this command.
+     *
+     * @return The flag argument for the command, or null if no flags are supported.
```
jpenilla, line 349: Document that it can return null in the main doc, for return something simple like `flag argument or null` is enough

cloud-core/src/main/java/cloud/commandframework/CommandManager.java
```diff
@@ -1417,7 +1417,16 @@ public enum ManagerSettings {
          * @since 1.2.0
          */
         @API(status = API.Status.STABLE, since = "1.2.0")
-        OVERRIDE_EXISTING_COMMANDS
+        OVERRIDE_EXISTING_COMMANDS,
+
+        /**
+         * When enabled, optional flag arguments are appended between each command node.
+         * It can have some conflicts when integrating with other command systems like brigadier.
```
jpenilla, line 1424:
```suggestion
         * It can have some conflicts when integrating with other command systems like Brigadier,
         * and code inspecting the command tree may need to be adjusted.
```

cloud-core/src/main/java/cloud/commandframework/CommandManager.java
```diff
@@ -1417,7 +1417,16 @@ public enum ManagerSettings {
          * @since 1.2.0
          */
         @API(status = API.Status.STABLE, since = "1.2.0")
-        OVERRIDE_EXISTING_COMMANDS
+        OVERRIDE_EXISTING_COMMANDS,
+
+        /**
+         * When enabled, optional flag arguments are appended between each command node.
```
jpenilla, line 1423:
```suggestion
         * Allows parsing flags at any position by appending flag argument nodes between each command node.
```

cloud-core/src/main/java/cloud/commandframework/arguments/compound/FlagArgument.java
```diff
@@ -79,6 +79,14 @@
      */
     public static final CloudKey<String> FLAG_META_KEY = SimpleCloudKey.of("__last_flag__", TypeToken.get(String.class));
 
+    /**
+     * Meta data for the set of parsed flags, used to detect duplicates
```
jpenilla, line 83:
```suggestion
     * Meta data for the set of parsed flags, used to detect duplicates.
```

cloud-core/src/main/java/cloud/commandframework/context/CommandContext.java
```diff
@@ -558,6 +559,24 @@ public void remove(final @NonNull CloudKey<?> key) {
         return this.getOptional(key).orElseGet(defaultSupplier);
     }
 
+    /**
+     * Get a value if it exists, else compute and store the value by the supplier and return it
+     *
+     * @param key             Argument key
+     * @param defaultSupplier Supplier of default value
```
jpenilla, line 566:
```suggestion
     * @param defaultFunction Default value function
```

cloud-core/src/main/java/cloud/commandframework/context/CommandContext.java
```diff
@@ -558,6 +559,24 @@ public void remove(final @NonNull CloudKey<?> key) {
         return this.getOptional(key).orElseGet(defaultSupplier);
     }
 
+    /**
+     * Get a value if it exists, else compute and store the value by the supplier and return it
+     *
+     * @param key             Argument key
```
jpenilla, line 565:
```suggestion
     * @param key             Cloud key
```

cloud-core/src/main/java/cloud/commandframework/context/CommandContext.java
```diff
@@ -558,6 +559,24 @@ public void remove(final @NonNull CloudKey<?> key) {
...
+     * @param defaultSupplier Supplier of default value
+     * @param <T>             Argument type
```
jpenilla, line 567:
```suggestion
     * @param <T>             Value type
```

cloud-core/src/main/java/cloud/commandframework/context/CommandContext.java
```diff
@@ -558,6 +559,24 @@ public void remove(final @NonNull CloudKey<?> key) {
...
+     * @return Argument, or supplied default value
```
jpenilla, line 568:
```suggestion
     * @return present or computed value
```

cloud-core/src/main/java/cloud/commandframework/context/CommandContext.java
```diff
@@ -558,6 +559,24 @@ public void remove(final @NonNull CloudKey<?> key) {
...
+     * @since 1.8.0
+     */
+    public <T> T computeIfAbsent(
+            final @NonNull CloudKey<T> key,
+            final @NonNull Function<CloudKey<T>, T> defaultSupplier
```
jpenilla, line 573:
```suggestion
            final @NonNull Function<CloudKey<T>, T> defaultFunction
```

jpenilla (changes requested):

cloud-core/src/main/java/cloud/commandframework/Command.java
jpenilla: This sounds reasonable to me

cloud-core/src/main/java/cloud/commandframework/context/CommandContext.java
Pablete1234: This follows all other methods like it (eg: getOrSupplyDefault) so i'll leave it as-is

cloud-core/src/main/java/cloud/commandframework/context/CommandContext.java
Pablete1234: This follows all other methods like it (eg: getOrSupplyDefault) so i'll leave it as-is

cloud-core/src/main/java/cloud/commandframework/context/CommandContext.java
jpenilla: The context stores more than just parser results, the old methods having outdated/wrong docs doesn't mean we should copy it over to new methods, at some point when I get to fixing the nullability annotations I will also fix this on old methods

cloud-core/src/main/java/cloud/commandframework/CommandManager.java
jpenilla: Looks like the second part got missed here

Pablete1234: my bad

jpenilla (approved):

Citymonstret (approved):

PR #396, opened 2022-10-18
Make CommandMethod inheritable

Phinner:
It would be very helpful for command systems that acts on polymorfic objects.

Example: Permissions where you have users and groups. Both have permissions, names, etc... but a group may have a weight for permission resolution and the users may have an id.

This PR aims to reduce the amount of code needed to accomplish that kind of task.

### DIFF
--- a/cloud-annotations/src/main/java/cloud/commandframework/annotations/AnnotationParser.java
+++ b/cloud-annotations/src/main/java/cloud/commandframework/annotations/AnnotationParser.java
@@ -224,7 +224,18 @@ public final class AnnotationParser<C> {
     static @NonNull List<Method> getAllDeclaredMethods(final Class<?> clazz) {
         final List<Method> methods = new ArrayList<>();
         final Set<MethodSignature> signatures = new HashSet<>();
-        for (Class<?> current = clazz; current != null; current = current.getSuperclass()) {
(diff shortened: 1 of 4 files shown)


cloud-annotations/src/main/java/cloud/commandframework/annotations/AnnotationParser.java
```diff
@@ -414,7 +414,7 @@ public void stringProcessor(final @NonNull StringProcessor stringProcessor) {
         /* Then register all parsers */
         this.parseParsers(instance);
         /* Then construct commands from @CommandMethod annotated classes */
-        final Method[] methods = instance.getClass().getDeclaredMethods();
+        final Method[] methods = instance.getClass().getMethods();
```
derklaro, line 417: This will break all commands whose method is declared as private...

Phinner:
Ok, see the above commits. I fixed the problem you mentioned my using an intermediate class to store the method signatures so I am able to use `getDeclaredMethods` while climbing the class hierarchy, without having duplicates.

Phinner:
After some time and tinkering with this feature, I think it's not worth it. It's easier to create a annotation command for the shared commands and build the specific commands with a builder than keeping track of a class hierarchy of inherited methods.

PR #397, opened 2022-10-18
Fix argument annotation name not being processed

Phinner:
You get a funny NPE when you use processed arguments names. Lemme fix that :)


Phinner:
Mmmh, now that I'm on it, not every command annotation value is processed...

PR #399, opened 2022-10-26
Rewrite Bukkit entity selector arguments using WrappedBrigadierParser instead of Bukkit API

jpenilla:
- Proper handling of spaces
- Possible to use built-in minecraft translations for brigadier exceptions
- Fixes suggestions on Paper in combination with their option to fix tag selector suggestions
- Added option to fail parse when the result collection is empty

Tested to work on 1.8.8, 1.13.2, 1.16.5, 1.19.2 Paper


Citymonstret (approved):
lgtm

PR #400, opened 2022-10-31
Fix regex exception occuring on some prefixes in the javacord module

Phinner:
An exception occurs if you use a regex meta character as a prefix in the javacord implementation. This fix uses the trick I found in the JDA implementation where the prefix length is used to cut the command input.


Phinner:
Yay, 400th PR :champagne:

Citymonstret (approved):

jpenilla (approved):

PR #401, opened 2022-11-01
fix invalid suggestions when parsing of argument fails

derklaro:
The current suggestion provider for dynamic arguments adds the suggestions of a failed parsed argument as well when stepping over it. This can for example be triggered when an integer argument is required, but a wrong value is provided. Consider the following case: `some_command <number <enum>`. When the user inputs something like `some_command world ` and tries to tab complete, the suggestions for the number argument show up. This is caused by a failed parse of the number argument, which doesn't step out of the current suggestion process for the argument, but rather continues to add all suggestions for it.

To fix this behaviour, we need to break the completion process if either the preprocessor or the parser isn't accepting the input and more arguments are provided.

### DIFF
--- a/cloud-core/src/test/java/cloud/commandframework/CommandSuggestionsTest.java
+++ b/cloud-core/src/test/java/cloud/commandframework/CommandSuggestionsTest.java
@@ -24,6 +24,7 @@
 package cloud.commandframework;
 
 import cloud.commandframework.arguments.compound.ArgumentTriplet;
+import cloud.commandframework.arguments.parser.ArgumentParseResult;
 import cloud.commandframework.arguments.standard.BooleanArgument;
 import cloud.commandframework.arguments.standard.EnumArgument;
 import cloud.commandframework.arguments.standard.IntegerArgument;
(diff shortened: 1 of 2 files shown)


cloud-core/src/test/java/cloud/commandframework/CommandSuggestionsTest.java
```diff
@@ -390,7 +404,7 @@ void testLiteralWithVariable() {
         Assertions.assertEquals(Arrays.asList("vici", "vidi"), suggestions3);
         final String input4 = "literal_with_variable vidi";
         final List<String> suggestions4 = manager.suggest(new TestCommandSender(), input4);
-        Assertions.assertEquals(Collections.emptyList(), suggestions4);
+        Assertions.assertEquals(Collections.singletonList("vidi"), suggestions4);
```
derklaro, line 407: This test was disabled & wrong previously. Without my change the test wouldn't pass as well as "vidi" would be given as a suggestion.

derklaro:
One thing I just tough of is that with this change the suggestion processor now pretty much relies on the fact that people are using the api correctly. While there *shouldn't* be any changes made to the command input queue during the (pre-) process, the whole check & further argument tries will fail *if* users are changing the queue.

I will change that later, so that the queue gets reset in case of a (pre-) processing failure.

derklaro:
Valid argument:
image

Invalid argument:
image

Citymonstret (approved):

jpenilla (approved):

PR #402, opened 2022-11-06
Make injection order predictable

Pablete1234:
Very simple change that makes it so the order injectors are called in is the one they were registered in, instead of being arbitrary.

This is relevant when the injectors overlap on classes, eg: CommandSender vs Player injectors, if your command wants to be injected a "CommandSender", it could randomly first attempt the player one (as player is compatible) and if player isn't available (eg: console) then it would go to the other injector for the actual CommandSender.

While this is probably a minor change it otherwise becomes impossible to make required injectors that will throw exceptions (later handled by the exception handler) when the necessary context isn't available when classes overlap. In the previous example, if the injector for Player threw an exception when the caller isn't a player, then randomly no command would be able to run from console (if you're unlucky enough that the player injector is called first).

With this change, as long as you register CommandSender's injector first, and then Player, such behavior works reliably.

### DIFF
--- a/cloud-core/src/main/java/cloud/commandframework/annotations/injection/ParameterInjectorRegistry.java
+++ b/cloud-core/src/main/java/cloud/commandframework/annotations/injection/ParameterInjectorRegistry.java
@@ -31,7 +31,7 @@ import io.leangen.geantyref.TypeToken;
 import java.util.ArrayList;
 import java.util.Collection;
 import java.util.Collections;
-import java.util.HashMap;
+import java.util.LinkedHashMap;
 import java.util.LinkedList;
 import java.util.List;
 import java.util.Map;
@@ -51,7 +51,7 @@ import org.checkerframework.checker.nullness.qual.Nullable;
 public final class ParameterInjectorRegistry<C> implements InjectionService<C> {
 
     private volatile int injectorCount = 0;
-    private final Map<Class<?>, List<ParameterInjector<C, ?>>> injectors = new HashMap<>();
+    private final Map<Class<?>, List<ParameterInjector<C, ?>>> injectors = new LinkedHashMap<>();
     private final ServicePipeline servicePipeline = ServicePipeline.builder().build();
 
     /**
--- a/cloud-core/src/main/java/cloud/commandframework/annotations/injection/ParameterInjectorRegistry.java
+++ b/cloud-core/src/main/java/cloud/commandframework/annotations/injection/ParameterInjectorRegistry.java
@@ -64,7 +64,7 @@ public final class ParameterInjectorRegistry<C> implements InjectionService<C> {
     }
 
     /**
-     * Register an injector for a particular type or any of it's assignable supertypes
+     * Register an injector for a particular type or any of it's assignable supertypes.
      *
      * @param clazz    Type that the injector should inject for. This type will matched using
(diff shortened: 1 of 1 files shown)


cloud-core/src/main/java/cloud/commandframework/annotations/injection/ParameterInjectorRegistry.java
```diff
@@ -63,7 +64,7 @@ public ParameterInjectorRegistry() {
     }
 
     /**
-     * Register an injector for a particular type
+     * Register an injector for a particular type or any of it's assignable supertypes
```
jpenilla, line 67:
```suggestion
     * Register an injector for a particular type or any of it's assignable supertypes.
```

cloud-core/src/main/java/cloud/commandframework/annotations/injection/ParameterInjectorRegistry.java
```diff
@@ -74,8 +75,21 @@ public synchronized <T> void registerInjector(
             final @NonNull Class<T> clazz,
             final @NonNull ParameterInjector<C, T> injector
     ) {
-        this.injectors.computeIfAbsent(clazz, missingClass -> new LinkedList<>()).add(injector);
-        this.injectorCount++;
+        this.registerInjector(cl -> clazz.isAssignableFrom(cl), injector);
+    }
+
+    /**
+     * Register an injector for a particular type predicate
```
jpenilla, line 82:
```suggestion
     * Register an injector for a particular type predicate.
```

cloud-core/src/main/java/cloud/commandframework/annotations/injection/ParameterInjectorRegistry.java
```diff
@@ -74,8 +75,21 @@ public synchronized <T> void registerInjector(
...
+     *
+     * @param predicate A predicate that matches if the injector should be used for a type.
+     * @param injector The injector that should inject the value into the command method
+     * @param <T>      Injected type
+     */
```
jpenilla, line 87: missing `@since` in jd and `@API` annotation, can use stable status for this

cloud-core/src/main/java/cloud/commandframework/annotations/injection/ParameterInjectorRegistry.java
```diff
@@ -92,13 +106,10 @@ public synchronized <T> void registerInjector(
...
-        final List<@NonNull ParameterInjector<C, ?>> injectors = new ArrayList<>(this.injectorCount);
-        for (final Map.Entry<Class<?>, List<ParameterInjector<C, ?>>> entry : this.injectors.entrySet()) {
-            if (clazz.isAssignableFrom(entry.getKey())) {
-                injectors.addAll(entry.getValue());
-            }
-        }
-        return Collections.unmodifiableCollection(injectors);
+        return Collections.unmodifiableCollection(this.injectors.stream()
+                .filter(pair -> pair.getFirst().test(clazz))
+                .map(pair -> pair.getSecond())
```
jpenilla, line 111: Pair::getSecond

jpenilla (changes requested):

jpenilla (approved):

PR #405, opened 2022-11-20
Improve nullability annotations on generics in CommandContext

jpenilla:
Allows this to work properly without `!!`
```kotlin
val a: String = ctx.getOrDefault("a", "s")
val b: String? = ctx.getOrDefault("b", null)
val c: String? = ctx.getOrSupplyDefault("c") { null }
val d: String = ctx.getOrSupplyDefault("d") { "s" }
val e = ctx.getOrDefault("e", "s") // will infer String!
val f = ctx.getOrDefault("f", null as String?) // will infer String?
val g = ctx.getOrDefault("g", null) // will infer Nothing?
```

Intellij doesn't seem as smart when it comes to Java, but I think the new behavior is still better than before where it would yell at you that it can be null when the default value is non-null.

### DIFF
--- a/cloud-core/src/main/java/cloud/commandframework/context/CommandContext.java
+++ b/cloud-core/src/main/java/cloud/commandframework/context/CommandContext.java
@@ -200,7 +200,7 @@ public class CommandContext<C> {
      * @param value Value
      * @param <T>   Value type
      */
-    public <T> void store(final @NonNull String key, final @NonNull T value) {
+    public <T extends @NonNull Object> void store(final @NonNull String key, final T value) {
         this.internalStorage.put(SimpleCloudKey.of(key), value);
     }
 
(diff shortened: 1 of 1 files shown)


solonovamax:
for all the `getOrDefault`  (and similar) functions, I believe that the signature should be:

```java
public <T> @PolyNull T getOrDefault(
            final @NonNull CloudKey<@NonNull T> key,
            final @PolyNull T defaultValue
    )
```

this is equivalent to specifying
```java
public <T> @NonNull T getOrDefault(
            final @NonNull CloudKey<@NonNull T> key,
            final @NonNull T defaultValue
    )
```
and
```java
public <T> @Nullable T getOrDefault(
            final @NonNull CloudKey<@NonNull T> key,
            final @Nullable T defaultValue
    )
```

according to the checher framework's docs, the `@PolyNull` annotation is equivalent to having two versions of the method; one with all the `@PolyNull` annotations replaced with `@NonNull`, and another with them all replaced with `@Nullable`

alternatively, there is also guava's `@ParametricNullness`, which I believe can be used to similar effects.

jpenilla:
IntelliJ doesn't properly handle PolyNull, it only properly handles JB's equivalent, however we don't use JB, what was done works as intended even though it's not pretty

jpenilla:
Yeah, I really wish it did, should probably open an issue somewhere on one of JetBrains trackers

PR #410, opened 2022-11-26
Improve FilteringCommandSuggestionProcessor and adjust default filters

jpenilla:


cloud-core/src/main/java/cloud/commandframework/execution/FilteringCommandSuggestionProcessor.java
```diff
@@ -46,12 +98,19 @@
         if (context.getInputQueue().isEmpty()) {
             input = "";
         } else {
-            input = context.getInputQueue().peek();
+            input = String.join(" ", context.getInputQueue());
         }
+        final @Nullable String inputLower = this.ignoreCase ? input.toLowerCase(Locale.ENGLISH) : null;
```
Pablete1234, line 103: I believe both here and a bit below, what you're looking for is `toLowerCase(Locale.ROOT)`, for language-insensitive to-lower-case

jpenilla:
This is causing a test to fail (due to using remaining unconsumed input instead of queue head) but I think the expected result is wrong. It's expecting a suggestion of `hello` when the input is `hello -`, which shouldn't be expected when we filter with startsWith in the default processor. We should expect nothing to be suggested which is the new behavior.

I see this more as a fix, but it is technically a behavior change, so we *could* add an option for whether to test the queue head or the remaining input (default to head). But again I think the breaks in this case would be breaking broken behavior by replacing it with correct behavior.

jpenilla:
I don't think it's actually documented anywhere what the default suggestion processor will be, so it might be nice to change the default to either
```java
FilteringCommandSuggestionProcessor.Filter.<CommandSender>startsWith(true).andTrimBeforeLastSpace()
```
or
```java
FilteringCommandSuggestionProcessor.Filter.<CommandSender>contains(true).andTrimBeforeLastSpace()
```
?

Pablete1234:
I believe the change is minor enough that it should be fine. If anyone has an issue with it they can just add a one-liner to replace it after updating to 1.8.0.

jpenilla:
I made the default case insensitive, it might also make sense to `.andTrimBeforeLastSpace()` for the default on Minecraft platforms? Since we are testing the joined input now. If someone wanted to suggest single words at a time (which was broken before anyway as it checked the queue head not tail) they already need to override that, so they will be setting a custom filter/processor anyways.

Citymonstret (approved):

derklaro:
On another note (maybe not directly related): the CommandTree is filtering suggestions for literal arguments already: https://github.com/Incendo/cloud/blob/a53177ac86b6a103d9e94766f4c86e34dec7cfc2/cloud-core/src/main/java/cloud/commandframework/CommandTree.java#L558-L563
which feels a bit off when considering that there is actual api to filter matching suggestions, but you cannot filter these (for example if you want to match using string similarity)

jpenilla:
> On another note (maybe not directly related): the CommandTree is filtering suggestions for literal arguments already:
> 
> https://github.com/Incendo/cloud/blob/a53177ac86b6a103d9e94766f4c86e34dec7cfc2/cloud-core/src/main/java/cloud/commandframework/CommandTree.java#L558-L563
> 
> which feels a bit off when considering that there is actual api to filter matching suggestions, but you cannot filter these (for example if you want to match using string similarity)

I think it's out of scope for this PR, but also using the suggestion processor for literals sounds like a good change

PR #411, opened 2022-12-07
cloud-fabric: Update for Minecraft 1.19.3

jpenilla:


zml2008:
I'm not sure we want to wholesale delete the MobEffectArgument -- it's kinda an API break. What about deprecating it and having it use a `RegistryEntryArgument` parser instead?

jpenilla:
Sounds reasonable to me

PR #414, opened 2022-12-10
Allow greedy parsers to suggest after a space

### LINKED ISSUE #190: Greedy parsers cannot have functioning suggestions

Pablete1234:
Fixes #190

### DIFF
--- a/cloud-core/src/main/java/cloud/commandframework/CommandTree.java
+++ b/cloud-core/src/main/java/cloud/commandframework/CommandTree.java
@@ -630,9 +630,14 @@ public final class CommandTree<C> {
         if (commandQueue.isEmpty()) {
             return Collections.emptyList();
         } else if (child.isLeaf()) {
-            String input = commandQueue.size() == 1 ? commandQueue.peek()
-                    : child.getValue() instanceof CompoundArgument ? ((LinkedList<String>) commandQueue).getLast()
-                            : String.join(" ", commandQueue);
+            final String input;
+            if (commandQueue.size() == 1) {
+                input = commandQueue.peek();
+            } else {
+                input = child.getValue() instanceof CompoundArgument
+                        ? ((LinkedList<String>) commandQueue).getLast()
+                        : String.join(" ", commandQueue);
+            }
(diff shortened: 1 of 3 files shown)


cloud-core/src/main/java/cloud/commandframework/CommandTree.java
```diff
@@ -629,15 +629,11 @@ private CommandTree(final @NonNull CommandManager<C> commandManager) {
...
-        } else if (child.isLeaf() && commandQueue.size() < 2) {
-            return this.directSuggestions(commandContext, child, commandQueue.peek());
         } else if (child.isLeaf()) {
-            if (child.getValue() instanceof CompoundArgument) {
-                final String last = ((LinkedList<String>) commandQueue).getLast();
-                commandContext.setCurrentArgument(child.getValue());
-                return child.getValue().getSuggestionsProvider().apply(commandContext, last);
-            }
-            return Collections.emptyList();
+            String input = commandQueue.size() == 1 ? commandQueue.peek()
```
jpenilla, line 633: please make this final, and imo the double ternary is still readable but it might be better to have one of them be an if to make things more obvious

jpenilla (changes requested):

jpenilla (approved):

Citymonstret (approved):

PR #416, opened 2022-12-11
Fix invalid syntax message containing commands the sender is not permitted to use

jpenilla:


jpenilla:
Current just a working PoC, not sure if this is the exact approach we want to take in the end

PR #418, opened 2022-12-12
Support tooltips on suggestions

masmc05:
Support for 
image
This pr adds support for description to suggestions, while fully maintaining old string suggestions

TODO:

- [x] add support for platforms
- [x] decide how we should transform suggestions to components for minecraft (minimessage, json, legacy, plain text?)
- [x] Review javadocs
- [x] Fix totally wrong API annotations to new methods
- [x] maybe some name changing


masmc05:
Added examples on velocity and paper in how to use this new feature
Here are the screenshots of the result
Velocity:
image
Paper:
image

Also deprecated methods that may result in losing the tooltip because of completion -> string -> completion transformation due to lack of support
All internal parsers now won't lose the tooltip

Soon will rebase

cloud-core/src/main/java/cloud/commandframework/CommandManager.java
```diff
@@ -108,8 +108,7 @@
 
     private CaptionVariableReplacementHandler captionVariableReplacementHandler = new SimpleCaptionVariableReplacementHandler();
     private CommandSyntaxFormatter<C> commandSyntaxFormatter = new StandardCommandSyntaxFormatter<>();
-    private CommandSuggestionProcessor<C> commandSuggestionProcessor =
-            new FilteringCommandSuggestionProcessor<>(FilteringCommandSuggestionProcessor.Filter.startsWith(true));
+    private CommandCompletionProcessor<C> commandCompletitionProcessor = new FilteringCommandCompletionProcessor<>();
```
zml2008, line 111:
```suggestion
    private CommandCompletionProcessor<C> commandCompletionProcessor = new FilteringCommandCompletionProcessor<>();
```

cloud-minecraft/cloud-paper/src/main/java/cloud/commandframework/paper/AsyncCommandSuggestionsListener.java
```diff
@@ -36,8 +51,68 @@
...
     private final PaperCommandManager<C> paperCommandManager;
 
+    private final BiConsumer<AsyncTabCompleteEvent, List<Completion>> completionsApplier;
+
+    @SuppressWarnings("ConstantConditions")
     AsyncCommandSuggestionsListener(final @NonNull PaperCommandManager<C> paperCommandManager) {
         this.paperCommandManager = paperCommandManager;
+        if (paperCommandManager.hasCapability(CloudBukkitCapabilities.PAPER_TOOLTIPS)) {
+            BiFunction<String, Message, AsyncTabCompleteEvent.Completion> completionWithDescription;
+            if (Audience.class.isAssignableFrom(Player.class)) { //If we use native adventure
```
zml2008, line 61:
```suggestion
            if (Audience.class.isAssignableFrom(Player.class)) { // If we use native adventure
```

(also, is this ok for running when Paper doesn't have adventure present? I think the capability check already ensures that this only runs on post-adventure releases)

cloud-minecraft/cloud-paper/src/main/java/cloud/commandframework/paper/AsyncCommandSuggestionsListener.java
```diff
@@ -36,8 +51,68 @@
...
+                completionWithDescription = (s, desc) -> {
+                    Component component = PaperBrigadier.componentFromMessage(desc);
+                    return AsyncTabCompleteEvent.Completion.completion(s, component);
+                };
+            } else { //We have a shaded adventure, using method handles to get needed methods without shading
```
zml2008, line 66:
```suggestion
            } else { // We have a shaded adventure, using method handles to get needed methods without shading
```

cloud-minecraft/cloud-brigadier/src/main/java/cloud/commandframework/brigadier/CloudBrigadierManager.java
```diff
@@ -655,16 +656,21 @@ public void registerDefaultArgumentTypeSupplier(
...
 
-        for (final String suggestion : suggestions) {
-            String tooltip = argument.getName();
-            if (!(argument instanceof StaticArgument)) {
-                if (argument.isRequired()) {
-                    tooltip = '<' + tooltip + '>';
-                } else {
-                    tooltip = '[' + tooltip + ']';
+        for (final Completion completion : completions) {
+            if (completion instanceof NativeCompletion) {
```
zml2008, line 660: This is pretty clunky when working with Adventure `Component`s, since they don't implement Message at compile time. We should allow end users to inject custom handling here (and for paper/others? with platform-specific completion handling) -- to support adventure components/whatever else

cloud-core/src/main/java/cloud/commandframework/execution/CommandCompletionProcessor.java
```diff
@@ -0,0 +1,49 @@
...
+import cloud.commandframework.Completion;
+import cloud.commandframework.execution.preprocessor.CommandPreprocessingContext;
+import java.util.List;
+import java.util.function.BiFunction;
+import org.checkerframework.checker.nullness.qual.NonNull;
+
+/**
+ * Processor that formats command suggestions
+ *
+ * @param <C> Command sender type
```
zml2008, line 35:
```suggestion
 * Processor that formats command completions.
 *
 * @param <C> Command sender type
 * @since 1.9.0
```

cloud-core/src/main/java/cloud/commandframework/CommandManager.java
```diff
@@ -218,6 +217,27 @@ protected CommandManager(
         );
         return this.commandSuggestionEngine.getSuggestions(context, input);
     }
+    /**
+     * Get command suggestions for the "next" argument that would yield a correctly parsing command input. The command
+     * completions provided by the command argument parsers will be filtered using the {@link CommandCompletionProcessor}
+     * before being returned.
```
zml2008, line 223:
```suggestion
     * Get command suggestions for the "next" argument that would yield a correctly parsing command input. 
     *
     * <p>The command completions provided by the command argument parsers will be filtered using 
     * the {@link CommandCompletionProcessor} before being returned.</p>
```

zml2008 (changes requested):
A few design comments -- you'll definitely need to do a javadoc pass to ensure that your changes have (correct) `@since` tags, and follow proper formatting with a summary fragment for Javadoc text.

cloud-minecraft/cloud-paper/src/main/java/cloud/commandframework/paper/AsyncCommandSuggestionsListener.java
masmc05: I couldn't find a version of AsyncTabCompleteEvent.Completion without adventure, as before it appeared it was using basic strings, and no support for tooltips

masmc05: so tooltip capability already means that we have adventure

cloud-minecraft/cloud-brigadier/src/main/java/cloud/commandframework/brigadier/CloudBrigadierManager.java
masmc05: also only on velocity we can support adventure directly, paper should work without (for versions before 1.16.5), bukkit without (bukkit never had adventure), fabric without, and i can't put such utilities to extras as on every platform the adventure needs to be handled differently

cloud-minecraft/cloud-brigadier/src/main/java/cloud/commandframework/brigadier/NativeCompletion.java
```diff
@@ -0,0 +1,116 @@
...
+    static @NonNull Completion ofUnknown(@NonNull final String suggestion, @Nullable final Object description) {
+        Message message = null;
+        if (description instanceof Message) {
+            message = (Message) description;
+        }
+        return message == null ? Completion.of(suggestion) : new SimpleCompletion(suggestion, message);
+    }
+
+    @API(status = API.Status.INTERNAL)
+    final class SimpleCompletion implements NativeCompletion {
```
zml2008, line 86: This shouldn't be public, just put it into another class

cloud-minecraft/cloud-brigadier/src/main/java/cloud/commandframework/brigadier/NativeCompletion.java
```diff
@@ -0,0 +1,116 @@
...
+        return new SimpleCompletion(suggestion, description);
+    }
+    /**
+     * Gives a suggestion instance with native brigadier message as description
+     * if the description object is a brigadier message
+     * @param suggestion the suggestion
+     * @param description the rich text as description
+     * @return a new suggestion instance
+     */
+    static @NonNull Completion ofUnknown(@NonNull final String suggestion, @Nullable final Object description) {
```
zml2008, line 77: no way, this is nasty

cloud-minecraft/cloud-brigadier/src/main/java/cloud/commandframework/brigadier/CloudBrigadierManager.java
zml2008: The API we expose for end users *must* include: `RichCompletion.of(String value, @Nullable Component tooltip)`. We can expose a settable function on the `CloudBrigadierManager` to perform the mapping.

Your workaround in the bukkit example is ugly af and users shouldn't have to deal with that directly -- especially if they're doing a cross-platform project, we don't want them to be forced to write their own abstraction just to produce completions with tooltips. Native platforms can set that up automatically. We could also have a fallback in `minecraft-extras` checking if the `Component` is an instanceof `Message` to have automatic handling for places where we we can do runtime class transformations (fabric, eventually forge).

zml2008 (changes requested):

cloud-minecraft/cloud-brigadier/src/main/java/cloud/commandframework/brigadier/NativeCompletion.java
masmc05: i thought this to be a way if they don't have brigadier dependency, but better I'll think of a way to add support for adventure for such reasons

cloud-minecraft/cloud-brigadier/src/main/java/cloud/commandframework/brigadier/CloudBrigadierManager.java
masmc05: did you mean something like https://github.com/Incendo/cloud/pull/418/commits/bb4b9cd41c19b5049cce9592604bb6ed771c6106?

masmc05:
idk, maybe I shouldn't provide explicit support for components here, but expose an api to registers a custom suggestion that supports that? 
the code becomes a mess and a spaghetti code when i try to add support for adventure when it seems like adventure shouldn't even be mentioned there

i can easily provide support for plain text as tooltips, but that's the maximum without making a mess

masmc05:
i think for most of cases plain text will be enough to make cool tooltips, so don't think that many projects would care about component tooltips

masmc05:
Is it good now or should I change something?

masmc05:
if you want to use this support as much as me, I posted it on `https://repo.masmc05.dev/repository/maven-snapshots/`, the version is `1.9.0-SNAPSHOT`, after this will be merged I'll delete it from there

Syrent:
I've tested this feature on Paper 1.20.1 (build 156) and it seems to not working correctly:
image


But it works perfectly fine on Paper 1.16.5 (build 794):
image

Argument code:
```kotlin
.argument(StringArgument.builder<ISender?>("mail")
    .withCompletionsProvider { context, _ ->
        val player = context.sender.player() ?: return@withCompletionsProvider emptyList()
        Database.getReceivedMails(player).filter { it.sender != null }.map { DescriptiveCompletion.of(it.id.toString(), "Example description") }
    }
)
```

masmc05:
> I've tested this feature on Paper 1.20.1 (build 156) and it seems to not working correctly: image
> 
> But it works perfectly fine on Paper 1.16.5 (build 794): image
> 
> Argument code:
> 
> ```kotlin
> .argument(StringArgument.builder<ISender?>("mail")
>     .withCompletionsProvider { context, _ ->
>         val player = context.sender.player() ?: return@withCompletionsProvider emptyList()
>         Database.getReceivedMails(player).filter { it.sender != null }.map { DescriptiveCompletion.of(it.id.toString(), "Example description") }
>     }
> )
> ```

Could not replicate this

Is the code 100% the same on both versions? Could you try forcing the registering of async command competition? If yes, did it change anything?

Syrent:
> > I've tested this feature on Paper 1.20.1 (build 156) and it seems to not working correctly: image
> > But it works perfectly fine on Paper 1.16.5 (build 794): image
> > Argument code:
> > ```kotlin
> > .argument(StringArgument.builder<ISender?>("mail")
> >     .withCompletionsProvider { context, _ ->
> >         val player = context.sender.player() ?: return@withCompletionsProvider emptyList()
> >         Database.getReceivedMails(player).filter { it.sender != null }.map { DescriptiveCompletion.of(it.id.toString(), "Example description") }
> >     }
> > )
> > ```
> 
> Could not replicate this
> 
> Is the code 100% the same on both versions? Could you try forcing the registering of async command competition? If yes, did it change anything?

Using `PaperCommandManager#registerAsynchronousCompletions` solved the problem

masmc05:
closing in favor of https://github.com/Incendo/cloud/pull/495

PR #422, opened 2022-12-13
Add command context type lookup methods

jpenilla:
Things to consider:
- Do we want a custom result type instead of `Pair`/`Collection<Pair>`?
- Overloads with default values for the two booleans (or some sort of `Query` with a builder? could be more powerful (add key filtering/type predicates?) but also more overkill)
- Adding tests

### DIFF
--- a/cloud-core/src/main/java/cloud/commandframework/context/CommandContext.java
+++ b/cloud-core/src/main/java/cloud/commandframework/context/CommandContext.java
@@ -36,10 +36,15 @@ import cloud.commandframework.keys.CloudKey;
 import cloud.commandframework.keys.CloudKeyHolder;
 import cloud.commandframework.keys.SimpleCloudKey;
 import cloud.commandframework.permission.CommandPermission;
+import cloud.commandframework.types.tuples.Pair;
+import java.util.ArrayList;
+import java.util.Collection;
 import java.util.Collections;
 import java.util.HashMap;
 import java.util.LinkedList;
+import java.util.List;
 import java.util.Map;
+import java.util.NoSuchElementException;
 import java.util.Optional;
 import java.util.function.Function;
 import java.util.function.Supplier;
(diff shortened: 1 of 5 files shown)


zml2008:
What was your idea of a use case on this? (i also only see one boolean?)

broccolai:
Not sure about this, the implementation looks good but I'm not sure we should be encouraging this kind of abuse with the command context.

cloud-core/src/main/java/cloud/commandframework/context/CommandContext.java
```diff
@@ -578,6 +583,112 @@ public <T> T computeIfAbsent(
...
+    }
+
+    /**
+     * Locates an element in the command context by type. If no matching elements
+     * are found, returns {@code null}.
+     *
+     * <p>If you have or are able to get the key for a value, it should be preferred
+     * to retrieve values by their key.</p>
+     *
+     * <p>When {@code exactType} is {@code flase}, a value may be returned
```
Citymonstret, line 631:
```suggestion
     * <p>When {@code exactType} is {@code false}, a value may be returned
```

cloud-core/src/main/java/cloud/commandframework/context/CommandContext.java
```diff
@@ -578,6 +583,112 @@ public <T> T computeIfAbsent(
...
+     * if it's type is assignable to {@code type}. Otherwise the type must pass
```
Citymonstret, line 632:
```suggestion
     * if its type is assignable to {@code type}. Otherwise the type must pass
```

Citymonstret (changes requested):
(As mentioned in Discord) I'd like to see the main signatures with TypeToken and have the existing signatures delegate to those, as CloudKey is TypeToken based. It'd be neat to compare those directly, rather than getting the underlying class

jpenilla:
I don't think this is something we want, injectors can be used for this purpose.

PR #423, opened 2022-12-14
Improve doc on ParameterInjectorRegistry#registerInjector

jpenilla:


Citymonstret (approved):

cloud-core/src/main/java/cloud/commandframework/annotations/injection/ParameterInjectorRegistry.java
```diff
@@ -80,7 +80,11 @@ public synchronized <T> void registerInjector(
     }
 
     /**
-     * Register an injector for a particular type predicate.
+     * Register an injector for a particular type predicate. The predicate should only
+     * return true if the injected type is assignable to the tested type. This predicate overload
+     * is provided in addition to {@link #registerInjector(Class, ParameterInjector)} to allow
+     * for exact, non-exact, or custom predicates, however is still bound by the aforementioned constraint.
+     * Failure to adhere to this will result in runtime exceptions.
```
zml2008, line 87:
```suggestion
     * Register an injector for a particular type predicate. 
     *
     * <p>The predicate should only
     * return true if the injected type is assignable to the tested type. This predicate overload
     * is provided in addition to {@link #registerInjector(Class, ParameterInjector)} to allow
     * for exact, non-exact, or custom predicates, however is still bound by the aforementioned constraint.
     * Failure to adhere to this will result in runtime exceptions.</p>
```

zml2008 (changes requested):

PR #440, opened 2023-04-19
add injection service support for suggestions methods.

### LINKED ISSUE #439: Add InjectionService support for @Suggestions methods.
As it is right now, @Suggestions methods are only able to handle 2 strict parameter types which are CommandContext and String(for input). Make it so that we will be able to use e.g. Guice injection services on these methods.
from this:
```java
@Inject
CageAddonManager addonManager;

@NotNull
@Suggestions("addon-ids")
public List<String> addonIds(final CommandContext<CommandSender> context, final String input) {
  return Strings.copyPartialMatchesSorted(
    input,
    this.addonManager.getPlugins().stream().map(PluginWrapper::getPluginId).toList()
  );
}
```
to this:
```java
@NotNull
@Suggestions("addon-ids")
public List<String> addonIds(final CageAddonManager addonManager/* guice */, final CommandSender sender /* MethodSuggestionsProvider */, final CommandContext<CommandSender> context /* ParameterInjectorRegistry */, final String input /* MethodSuggestionsProvider */) {
  return Strings.copyPartialMatchesSorted(
    input,
    addonManager.getPlugins().stream().map(PluginWrapper::getPluginId).toList()
  );
}
```

portlek:
closes #439


portlek:
The only concern I have is that since tab-completes are running more than a regular command execution, this might create some performance issues. so I thought I might add a boolean parameter for @Suggestions annotations that refer to this feature to be enabled. What do you think?

jpenilla:
> The only concern I have is that since tab-completes are running more than a regular command execution, this might create some performance issues. so I thought I might add a boolean parameter for @Suggestions annotations that refer to this feature to be enabled. What do you think?

You can check if the method signature takes exactly one command context and one string in that order. If so the old logic can be used.

portlek:
I rebase the PR to 1.9.0-dev again. it's ready to merge.

portlek:
@jpenilla could you review the PR, please?

Citymonstret:
This has been addressed in v2.

PR #443, opened 2023-05-07
fixed null permission

NonSwag:
minecraft (bukkit) commands registered by cloud have no permission
image
(not a good representation, but the two messages below shall represent the issue)


jpenilla:
`permission` is allowed to be null, as you can see by the null checks throughout Bukkit's `Command` class. Cloud commands do not necessarily have a string permission, which is why we override `testPermissionSilent` instead of setting the `permission` field.

NonSwag:
I know but `testPermissionSilent` flags as `false` in my case, but the permission is not set
And this fixes the issue.

PR #444, opened 2023-05-14
added permission when owner is null and command permission is not empty

NonSwag:
This pull request aims to fix the issue from my last pr #443
I realized what @jpenilla really meant with:
> [`permission` is allowed to be null](https://github.com/Incendo/cloud/pull/443#issuecomment-1537522935)

I added a condition, so the permission can still be null in case the permission is empty


NonSwag:
Hello?

NonSwag:
pump

PR #446, opened 2023-06-02
make StringArgument$Builder#withMode public

Machine-Maker:
Pretty sure this method is supposed to be public as it has a javadoc


Citymonstret:
I'll close this as it not relevant for Cloud v2, and this is not going to make it into any patch release for v1.

Issue #398, opened 2022-10-23
Add ComponentArgument to cloud-minecraft-extras


gepron1x:
add adventure component argument type, including quoted and greedy modes, similiar to StringArgument. plus provide support for different component serializers (json, legacy, miniMessage)

PR #447, opened 2023-06-02
add ComponentArgument

Machine-Maker:
Closes https://github.com/Incendo/cloud/issues/398

It's a little jank putting it in minecraft-extras, cause you can't register a brig mapping for it afaik.

You can do it a plugin easily with
```java
this.commands.brigadierManager().registerMapping(new TypeToken<ComponentArgument.Parser<CommandSender, ?>>() {}, builder -> builder
    .cloudSuggestions()
    .to(argument -> {
        return switch (argument.getStringMode()) {
            case QUOTED -> StringArgumentType.string();
            case GREEDY, GREEDY_FLAG_YIELDING -> StringArgumentType.greedyString();
            default -> StringArgumentType.word();
        };
    })
);
```

this also means you need a parser type, instead of just using `StringArgument$StringParser#map` to get a mapped parser.


broccolai:
Maybe we could include a static method for registering it on the ComponentArgument? Especially if users are always going to add it in the way you described

jpenilla:
This will need to be recreated for the https://github.com/Incendo/cloud-minecraft repo

PR #469, opened 2023-10-02
Expose CommandPermission in NoPermissionException

Konicai:
Not quite sure how I feel about the new method being fluent, while the other exceptions don't use fluent getters. This seemed like the best name though given what there is to work with.

### DIFF
--- a/cloud-core/src/main/java/cloud/commandframework/exceptions/NoPermissionException.java
+++ b/cloud-core/src/main/java/cloud/commandframework/exceptions/NoPermissionException.java
@@ -64,6 +64,9 @@ public class NoPermissionException extends CommandParseException {
     }
 
     /**
+     * Get the missing permission node
+     *
+     * @return Get the missing permission node
      * @deprecated use {@link #missingPermission()) instead
      */
     @API(status = API.Status.DEPRECATED, since = "1.8.5")
--- a/cloud-core/src/main/java/cloud/commandframework/exceptions/NoPermissionException.java
+++ b/cloud-core/src/main/java/cloud/commandframework/exceptions/NoPermissionException.java
@@ -64,12 +64,23 @@ public class NoPermissionException extends CommandParseException {
     }
 
     /**
-     * Get the missing permission node
-     *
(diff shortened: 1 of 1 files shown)


Konicai:
i believe i'll need to update the exception handlers of each platform to not use the deprecated method - i'll do so asap if this change is accepted

cloud-core/src/main/java/cloud/commandframework/exceptions/NoPermissionException.java
```diff
@@ -67,11 +67,25 @@ public final String getMessage() {
...
         return this.missingPermission.toString();
     }
 
+    /**
+     * Returns the missing {@link CommandPermission}
+     *
+     * @return the missing permission
+     * @since 1.8.5
+     */
+    @API(status = API.Status.STABLE, since = "1.8.5")
```
jpenilla, line 84: new API should target 1.9

Konicai: okay cool - `1.9.0` ?

Konicai: and the version for the deprecation annotation should also be `1.9.0` ?

jpenilla: yes and yes

cloud-core/src/main/java/cloud/commandframework/exceptions/NoPermissionException.java
```diff
@@ -67,11 +67,25 @@ public final String getMessage() {
...
+    @Deprecated
     public @NonNull String getMissingPermission() {
         return this.missingPermission.toString();
     }
 
+    /**
+     * Returns the missing {@link CommandPermission}
+     *
+     * @return the missing permission
+     * @since 1.8.5
```
jpenilla, line 82: missed a spot

jpenilla (approved):

PR #494, opened 2023-11-12
feature: command beans

Citymonstret:
This introduces a new way of registering commands that live in their own separate classes.

### DIFF
--- a/cloud-core/src/main/java/cloud/commandframework/CommandFactory.java
+++ b/cloud-core/src/main/java/cloud/commandframework/CommandFactory.java
@@ -1,0 +1,49 @@
+//
+// MIT License
+//
+// Copyright (c) 2022 Alexander Söderberg & Contributors
+//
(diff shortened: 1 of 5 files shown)


cloud-core/src/main/java/cloud/commandframework/CommandFactory.java
```diff
@@ -0,0 +1,48 @@
...
+    /**
+     * Creates a command using the given {@code commandManager}. Each invocation produces a unique instance of the command.
+     * <p>
+     * This method has no side effects, meaning that the created command will not be automatically registered to the
+     * {@link CommandManager command manager}.
+     *
+     * @param commandManager the command manager
+     * @return the created command
+     */
+    @NonNull Command<C> createCommand(@NonNull CommandManager<C> commandManager);
```
jpenilla, line 47: Maybe `List<Command<C>>`?

Issue #455, opened 2023-07-01
Default arguments aren't properly parsed


masmc05:
because 
https://github.com/Incendo/cloud/blob/a4cc749b91564af57bb7bba36dd8011b556c2b3a/cloud-core/src/main/java/cloud/commandframework/CommandTree.java#L346
don't split at the space, you can't give default arguments to complex types like bukkit locations, because they will see it always as 1 argument in the queue

Issue #500, opened 2023-11-14
Make default values more powerful


Citymonstret:
In Cloud v1 the default values are stored in the arguments, and the argument builder has the ability to serialize objects into strings. These values are then pushed onto the input queue when parsing.

In Cloud v2 the default values are stored in the command components, instead of the arguments. This means they bypass the argument entirely, and we thus lose the ability to pre-serialize these objects. However, instead of bringing back that functionality, I am instead proposing a new mechanism for default values.

```java
public interface DefaultValue<C, T> {

  @NonNull <T> evaluteDefault(@NonNull CommandContext<C> context);
}
```
We'd offer three types:
- constant: `DefaultValue.constant(5)`,  `DefaultValue.constant(Duration.ofSeconds(5L))`
- dynamic: `DefaultValue.dynamic(ctx -> ctx.sender().getLocation())`
- parsed: `DefaultValue.parsed("5")`, `DefaultValue.parsed("@p")`

constant/dynamic would bypass the optional argument entirely, whereas parsed would still push the default value onto the input

PR #505, opened 2023-11-15
feature: allow for dynamic default values

Citymonstret:
Implements #500.


masmc05:
can this also fix https://github.com/Incendo/cloud/issues/455 ?

Citymonstret:
> can this also fix #455 ?

I think this would get fixed automatically as of #498 rather than this PR, as it'd append to the input, spaces and all. With this PR you now have the option of supplying a location object as well if you're using builders. Either way, it should be resolved in 2.0.

PR #512, opened 2023-11-16
feature: add command sender to descriptions and flags

Citymonstret:
This is a re-implementation of #237. This allows us to show different descriptions per sender.

Command flags erased the sender types previously. We might as well fix that in 2.0.

This drops deprecated description methods.


Citymonstret:
This is still up in the air and might not be worth the hassle. The command flags will definitely have the command sender parameter added, but descriptions may not.

PR #513, opened 2023-11-17
build!: target java 17 

broccolai:
Could also target 11 instead, but seeing as this will be our only change to change target for a long time I thought we may as well go to 17.

1.8 servers etc can run just fine on Java 17 with the flag, is there still a reason to target Java 8?

This PR doesn't update the code to use new features introduced since Java 8, except from a [fix for #isAccessible](https://docs.oracle.com/en/java/javase/17/docs/api/java.base/java/lang/reflect/AccessibleObject.html#isAccessible())

### DIFF
--- a/build-logic/src/main/kotlin/cloud.base-conventions.gradle.kts
+++ b/build-logic/src/main/kotlin/cloud.base-conventions.gradle.kts
@@ -11,8 +11,8 @@ plugins {
 indra {
     javaVersions {
         minimumToolchain(17)
-        target(8)
-        testWith(8, 11, 17)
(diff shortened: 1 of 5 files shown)


solonovamax:
imo, go for java 11 if you really want to drop java 8 support

Citymonstret:
> imo, go for java 11 if you really want to drop java 8 support

We're going to stick to Java 8 unless there's a very compelling reason for us to force the version up. So far there has been none.

jpenilla:
We won't be bumping the Java requirement for core modules in 2.0

PR #514, opened 2023-11-17
chore(deps): update dependencies

broccolai:
should all be safe updates


jpenilla:
tests fail with j8: Caused by: java.lang.UnsupportedClassVersionError: org/mockito/junit/jupiter/MockitoExtension has been compiled by a more recent version of the Java Runtime (class file version 55.0), this version of the Java Runtime only recognizes class file versions up to 52.0

PR #521, opened 2023-11-20
feature: make parsing & suggestions return futures

Citymonstret:


cloud-core/src/main/java/cloud/commandframework/arguments/parser/ArgumentParser.java
```diff
@@ -153,4 +193,27 @@ default boolean isContextFree() {
...
+
+    @API(status = API.Status.STABLE, since = "2.0.0")
+    interface FutureArgumentParser<C, T> extends ArgumentParser<C, T> {
+
+        @Override
+        default @NonNull ArgumentParseResult<@NonNull T> parse(
+                @NonNull CommandContext<@NonNull C> commandContext,
+                @NonNull CommandInput commandInput
+        ) {
+            return this.parseFuture(commandContext, commandInput)
```
Citymonstret, line 206: this is not needed now that the suggestions use parseFuture

Citymonstret:
note for self: https://github.com/jhalterman/concurrentunit is cool and can also be used to test command methods by invoking the waiter from the command handler.

PR #532, opened 2023-11-30
Expose argument names to minecraft extras help message provider

jpenilla:
Supersedes #381


Citymonstret (approved):
This is a lot cleaner when not having to support the old bifunction :D

PR #536, opened 2023-12-02
feature(core): exception handling

Citymonstret:
This introduces a new exception handling system that allows us to register exception handlers for both the precise exception types _and_ their super types. This system also allows us to register multiple exception handlers for the same exception type.

Before we make use of this API in the platform implementations I want to make sure that it's refined. Feedback is very welcome :)

### DIFF
--- a/cloud-core/src/main/java/cloud/commandframework/exceptions/handling/ExceptionHandler.java
+++ b/cloud-core/src/main/java/cloud/commandframework/exceptions/handling/ExceptionHandler.java
@@ -1,0 +1,63 @@
+//
+// MIT License
+//
+// Copyright (c) 2022 Alexander Söderberg & Contributors
+//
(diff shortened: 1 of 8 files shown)


cloud-core/src/main/java/cloud/commandframework/exceptions/handling/ExceptionHandler.java
```diff
@@ -0,0 +1,150 @@
...
+            final Throwable cause = ctx.exception().getCause();
+            if (cause != null && predicate.test(cause)) {
+                throw cause;
+            }
+            throw ctx.exception();
+        };
+    }
+
+    /**
+     * Returns an exception handler that throws the cause of the {@link ExceptionContext#exception()} if it's and instance
```
Konicai, line 110: `and` instance

PR #542, opened 2023-12-03
Correct Fabric's string reader injection

zml2008:
This mixin had failed silently at some point.


jpenilla:
Do we have any idea how long ago this broke?

jpenilla:
We don't plan to make any more v1 releases, and 1.9.0-dev is known working, so I'm going to close this as I haven't tested it

PR #543, opened 2023-12-04
feature: new help system

Citymonstret:
this system aims to be more flexible and extensible than the old system, while behaving the same as the existing system

this doesn't really touch the logic used to generate the results, nor the MinecraftHelp logic for displaying the results
(although there are definitely things that can be improved there)

TODO:
- [x] Replace the deleted unit tests.
- [x] Make sure MinecraftHelp behaves the same as before.


cloud-core/src/main/java/cloud/commandframework/CommandManager.java
```diff
@@ -992,36 +995,59 @@ public final <E extends Exception> void handleException(
...
      *
      * @return a new command helper handler instance
-     * @since 1.7.0
+     * @since 2.0.0
      */
-    @API(status = API.Status.STABLE, since = "1.7.0")
-    public final @NonNull CommandHelpHandler<C> createCommandHelpHandler() {
-        return new CommandHelpHandler<>(this, cmd -> true);
+    @API(status = API.Status.STABLE, since = "2.0.0")
+    public final @NonNull HelpHandler<C> createCommandHelpHandler() {
```
powercasgamer, line 1006: Shouldn't this be renamed to `createHelpHandler`?

PR #545, opened 2023-12-05
Allow PredicatePermissions to return a PermissionResult

Konicai:
A few things:
- Will add a better PR description tomorrow, probably
- Will javadoc and fix checkstyle issues tomorrow, probably
- It seems like `AndPermission` and `OrPermission` had loose definitions of what to do when empty. `OrPermission` already resulted in false, but I added a check for `AndPermission` so that it now also results in false. I was thinking of requiring a non-empty Set in their constructors, and also moving the iteration to the respective classes. Can't move much more up without introducing a generic to CommandPermission, which seems yucky.
- I was originally going to make `PermissionLevelResult` with a private constructor and a static factory method. But then I figured that it wouldn't hurt to allow the class to be extended. So then I was looking at a protected constructor and a static factory method, and I figured to just have a public constructor. It feels like I'm getting bureaucratic. Maybe a middle ground?


cloud-minecraft/cloud-fabric/src/main/java/cloud/commandframework/fabric/FabricClientCommandManager.java
```diff
@@ -122,8 +123,10 @@ public boolean hasPermission(final @NonNull C sender, final @NonNull String perm
      * @since 1.5.0
      */
     public static <C> @NonNull PredicatePermission<C> integratedServerRunning() {
-        return sender -> Minecraft.getInstance().hasSingleplayerServer();
-    }
+        return PredicatePermission.of(
+                SimpleCloudKey.of("integrated-server-running"),
+                sender -> Minecraft.getInstance().hasSingleplayerServer()
+        );    }
```
Konicai, line 129: oop

powercasgamer (changes requested):

jpenilla:
as mentioned on discord, the result for an empty and permission shouldn't be changed. but requiring a non empty set also sounds reasonable.

Konicai:
thank you - got busy the last couple days. I'll be able to finish this today or tomorrow

Citymonstret:
> thank you - got busy the last couple days. I'll be able to finish this today or tomorrow

absolutely no rush, I just figured I'd help keep it up to date with 2.0.0-dev as it's still moving quite fast :sweat_smile:

Konicai:
@jpenilla can i get some guidance/clarification on the direction i should take with AndPermission/OrPermission :)

Citymonstret:
I merged in the changes from #578, hopefully I didn't butcher it too badly.

Citymonstret (approved):

Citymonstret (approved):

Konicai:
thank you... a bit taken by life right now 😆

jpenilla (approved):

PR #546, opened 2023-12-06
Remove legacy Brigadier adapter code

jpenilla:
CloudStringReader is a cheap wrapper, not sure if it's worth the effort to have similar mixins as before


zml2008 (approved):
might as well keep things simple for now, can always add mixin-implementations of the cloud interface on brig's StringReader if profiling results indicate it'd help

PR #549, opened 2023-12-06
chore: clean up CloudKey

Citymonstret:
This PR gets rid of SimpleCloudKey, and exposes static factory methods in CloudKey that reference an immutable implementation.

CloudKey is turned into an abstract class with a final implementation of equals (and hashCode) to enforce the requirement that only the names are checked for equality.

The prefixes are dropped from the CloudKey accessors.


jpenilla:
CloudKeyHolder#getKey could probably be renamed to cloudKey or just key

PR #551, opened 2023-12-08
feat(annotations): annotation-less arguments, and description mappers

Citymonstret:
```java
@CommandMethod("command <required> [optional] [optional_but_weird]")
public void command(TestCommandSender sender, String required, Integer optional, Boolean optionalButWeird) {
}
```


cloud-annotations/src/main/java/cloud/commandframework/annotations/processing/CommandMethodVisitor.java
```diff
@@ -177,6 +184,8 @@ public Void visitExecutable(final ExecutableElement e, final Void unused) {
                     );
                 }
             }
+
+
```
powercasgamer, line 188:
```suggestion
```

powercasgamer (changes requested):

PR #564, opened 2023-12-13
Update MinecraftExceptionHandler to better use new exception handling API

jpenilla:
The ExceptionType enum is a pointless restriction with the new API. Also renamed `with`-ers since they aren't used much in the rest of the codebase and look out of place. Also allows not sending a message by returning `null`.


Citymonstret (approved):
lgtm

PR #566, opened 2023-12-14
Preserve suggestion tooltips from WrappedBrigadierParsers, and don't join the Brig suggestions future

jpenilla:


Citymonstret (approved):
lgtm

PR #567, opened 2023-12-14
Clean up SuggestionProvider hierarchy

jpenilla:


Machine-Maker:
Any use for a static method on SuggestionProvider for some constant suggestions?

Also, a thought I had was about ArgumentParser not extending SuggestionProvider anymore. It just seems like it could be better designed than having ArgumentParser extend it but having implementations implementing a subinterface of SuggestionProvider as well.

What if ArgumentParser had a method that returned SuggestionParser? That way, the correct type of provider, blocking, nonblocking, strings, etc. would just be returned there instead of the overlapping interface implementations on separate types?

cloud-core/src/main/java/cloud/commandframework/arguments/parser/ArgumentParser.java
```diff
@@ -184,6 +141,23 @@ default int getRequestedArgumentCount() {
...
+     * Otherwise, {@link SuggestionProvider#noSuggestions()} will be returned.
+     *
+     * @return the suggestion provider
+     */
+    @SuppressWarnings("unchecked")
+    @Override
+    default @NonNull SuggestionProvider<C> suggestionProvider() {
+        if (this instanceof SuggestionProvider) {
+            return (SuggestionProvider<C>) this;
+        }
```
Machine-Maker, line 157: Ok, so what about just removing this? If ArgumentParsers are to default to no suggestions, simply implementing the SuggestionProvider interface shouldn't be enough to change behaviors, they should implement it and override to return `this`. I think making that explicit when creating a custom parser is better than just extending another interface on that parser.

jpenilla: I think my [comment](https://github.com/Incendo/cloud/pull/569#discussion_r1428421030) on #569 is applicable here too. To reiterate, I don't agree that it's better to require implementing extra methods to opt into suggestions without using a lambda, I think it's needless verbosity. The current code makes the majority of cases simpler to implement without impeding flexibility. What's the reasoning for why you think being "explicit" in this way would be better?

Machine-Maker: I wasn't suggesting removing the default implementation, but just the instanceof check.

It doesn't have to be explicit that you don't want any suggestions on the parser, but it should be explicit if you implement SuggestionProvider on the parser that you want to use the parser as the suggestion provider.

Citymonstret: > but it should be explicit if you implement SuggestionProvider on the parser that you want to use the parser as the suggestion provide

It is. You're explicitly stating your intent of providing suggestions by implementing the `SuggestionProvider` interface. By making your class implement an interface you inherit the contract of that interface. I do not think that it would make it clearer to anyone if we made the API more cumbersome to use by forcing them to implement another method re-stating this intent.

We do not want to force everyone to implement a suggestion method when creating a parser, because a parser doesn't necessarily have to produce suggestions. That is not the purpose of a parser. That can be true while we also acknowledge the fact that it is very likely for a parser to also provide suggestions. Because it's such a common usage of the API, it makes a lot of sense for us to make it easier and to get rid of any unnecessary barriers to easily achieving that.

broccolai (approved):
Haven't looked through all the parsers but LGTM once BlockingSuggestionProvider is un-subclass'd

PR #568, opened 2023-12-14
Refactor selector data types and exceptions

jpenilla:


cloud-minecraft/cloud-bukkit/src/main/java/cloud/commandframework/bukkit/parsers/selector/MultipleEntitySelectorParser.java
```diff
@@ -108,6 +111,16 @@ public MultipleEntitySelector mapResult(
         if (entities.isEmpty() && !this.allowEmpty) {
             new Thrower(NO_ENTITIES_EXCEPTION_TYPE.get()).throwIt();
         }
-        return new MultipleEntitySelector(input, entities);
+        return new MultipleEntitySelector() {
```
Machine-Maker, line 114: Is it not better to have a record implementation just for toString, equals, hashCode?

jpenilla: The old classes didn't implement those, and neither do the Fabric ones. In principle, I agree with you, but practically I don't think people will use these as map keys or compare them. Also, cloud is still using Java 8.

Citymonstret: You could just slap an `@ImmutableImpl` on the selector interface and use that. Less visible code :shrug:

jpenilla: Not really since that would be package private and we would have to expose factory methods that have no purpose for the end user

broccolai (approved):

PR #569, opened 2023-12-15
feat: don't make `ArgumentParser` implement `SuggestionProvider`

Citymonstret:


jpenilla:
Should remove all the `implements SuggestionProvider.NoSuggestions` - that interface can probably be removed and just be directly implemented as a lambda for `SuggestionProvider.noSuggestions` now.

cloud-core/src/main/java/cloud/commandframework/arguments/parser/ArgumentParser.java
```diff
@@ -139,6 +141,23 @@ default int getRequestedArgumentCount() {
...
+     *
+     * @return the suggestion provider
+     */
+    @SuppressWarnings("unchecked")
+    @Override
+    default @NonNull SuggestionProvider<C> suggestionProvider() {
+        if (this instanceof SuggestionProvider) {
+            return (SuggestionProvider<C>) this;
+        }
+        return SuggestionProvider.noSuggestions();
```
Machine-Maker, line 158: Would it be better if all all implementations of ArgumentParser could just implement it as returning this? I think that's better than the default implementation doing this. Parsers should always have suggestions and having a default impl doesn't force an implementation.

jpenilla: Parsers have always defaulted to no suggestions, they are parsers, not suggesters. They can provide fallback suggestions when no suggestion provider is explicitly set for a component. And no, I don't think it would be better to require implementing an extra method on every parser that wants to provide those fallback suggestions.

jpenilla (approved):

PR #570, opened 2023-12-15
chore: `List<Suggestion>` -> `Iterable<Suggestion>`

Citymonstret:


solonovamax:
why `Iterable` and not `Collection`?

that seems more appropriate to me

jpenilla:
Might as well change BlockingSuggestionProvider.Strings as well

jpenilla:
Being able to return an iterable is more flexible than a collection, but it's also a little more annoying to deal with on the consumer side...

broccolai:
Would prefer Iterable

PR #572, opened 2023-12-16
feat: make suggestion processors act on individual suggestions

Citymonstret:
We currently have to collect then process and then collect again, which is not ideal. By instead applying the filter when we can skip unwrapping suggestions only to immediately collect them again.


cloud-core/src/main/java/cloud/commandframework/execution/CommandSuggestionProcessor.java
```diff
@@ -25,19 +25,19 @@
...
-import java.util.function.BiFunction;
 import org.apiguardian.api.API;
 import org.checkerframework.checker.nullness.qual.NonNull;
+import org.checkerframework.checker.nullness.qual.Nullable;
 
 /**
  * Processor that formats command suggestions
  *
- * @param <C> Command sender type
+ * @param <C> the command sender type
```
jpenilla, line 35:
```suggestion
 * @param <C> command sender type
```

We should probably standardize this across the codebase... In general, I think "the" prefixes for params/return values don't add much value unless they're part of a longer statement.

cloud-core/src/main/java/cloud/commandframework/execution/CommandSuggestionProcessor.java
```diff
@@ -49,6 +49,17 @@ public interface CommandSuggestionProcessor<C> extends
      */
     @API(status = API.Status.STABLE, since = "1.8.0")
     static <C> @NonNull CommandSuggestionProcessor<C> passThrough() {
-        return (ctx, suggestions) -> suggestions;
+        return (ctx, suggestion) -> suggestion;
     }
+
+    /**
+     * Processes the given {@code suggestion} and returns the result.
+     * <p>
```
jpenilla, line 57: We should setup checkstyle to enforce closing tags :p

cloud-core/src/main/java/cloud/commandframework/execution/CommandSuggestionProcessor.java
```diff
@@ -49,6 +49,17 @@ public interface CommandSuggestionProcessor<C> extends
...
+     * If {@code null} is returned, the suggestion will be dropped.
+     *
+     * @param context    the context
+     * @param suggestion the suggestion
+     * @return the result
+     */
```
jpenilla, line 63: Probably want to add `@since`/`@API` here, or remove it from the above method

cloud-core/src/main/java/cloud/commandframework/execution/CommandSuggestionProcessor.java
```diff
@@ -49,6 +49,17 @@ public interface CommandSuggestionProcessor<C> extends
...
-        return (ctx, suggestions) -> suggestions;
+        return (ctx, suggestion) -> suggestion;
     }
+
+    /**
+     * Processes the given {@code suggestion} and returns the result.
+     * <p>
+     * If {@code null} is returned, the suggestion will be dropped.
+     *
+     * @param context    the context
```
jpenilla, line 60:
```suggestion
     * @param context    command preprocessing context
```

cloud-core/src/main/java/cloud/commandframework/execution/CommandSuggestionProcessor.java
```diff
@@ -49,6 +49,17 @@ public interface CommandSuggestionProcessor<C> extends
...
+     * @param suggestion the suggestion
```
jpenilla, line 61:
```suggestion
     * @param suggestion the suggestion to process
```

cloud-core/src/main/java/cloud/commandframework/execution/CommandSuggestionProcessor.java
```diff
@@ -49,6 +49,17 @@ public interface CommandSuggestionProcessor<C> extends
...
+     * @return the result
```
jpenilla, line 62:
```suggestion
     * @return processed suggestion or {@code null}
```

cloud-core/src/main/java/cloud/commandframework/execution/CommandSuggestionProcessor.java
Konicai: i came across this comment trying to figure out which i should follow 😂

Konicai: the only opinion i have is that having a `the` (or similar article) for `@return` tags is nice and flows well. if that is followed, then it should be for param tags too.

Citymonstret: > the only opinion i have is that having a the (or similar article) for @return tags is nice and flows well
I agree.

> then it should be for param tags too.
This I'm less concerned about. I agree that it'd be nice to be consistent, but I think it's more important that we're consistent within the same tag type.

PR #575, opened 2023-12-17
Remove cloud-tasks

jpenilla:
cloud-tasks was a nice idea but we've come to realize it doesn't fit that well in the project. Also, it's incompatible with Folia in its current form.

The removed documentation is no longer relevant with suggestions and parsers being able to return futures.


Citymonstret (approved):
lgtm

PR #577, opened 2023-12-18
feat: introduce SuggestionLike

Citymonstret:
This is **not** something we have decided to add to Cloud. This PR is here as a proof-of-concept so that we can discuss whether this is something we're interested in.

It has the downside of making the signatures even more complex. This could be solved by removing the ability to subtype and force all implementations to use `SuggestionLike` in the signature as well.

I personally do not feel like the *Like suffix is a great choice, but after a discussion in the Paper discord (naturally) we couldn't come up with a better one on the spot.


jpenilla:
I don't think this adds much value, and something similar could always be implemented as a wrapping suggestion provider.

PR #584, opened 2023-12-22
feat: make CF parsers return `ArgumentParseResult`

Citymonstret:


cloud-core/src/main/java/cloud/commandframework/arguments/aggregate/AggregateCommandParser.java
```diff
@@ -87,25 +88,34 @@ public interface AggregateCommandParser<C, O> extends ArgumentParser.FutureArgum
...
-                    future.thenCompose(result -> component.parser()
-                            .parseFuture(commandContext, commandInput)
-                            .whenComplete((value, exception) -> {
-                                if (value == null || exception != null) {
-                                    return;
-                                }
-                                final CloudKey key = CloudKey.of(component.name(), component.valueType());
-                                aggregateCommandContext.store(key, value);
-                            }));
+                    future.thenCompose(result -> {
```
Citymonstret, line 99: ew yucky bad pls someone suggest a better implementation

Citymonstret: we could just return CompletableFuture<O> from mappers?

jpenilla (approved):

PR #590, opened 2023-12-23
bukkit: parse selectors in correct thread context

jpenilla:


cloud-minecraft/cloud-bukkit/src/main/java/cloud/commandframework/bukkit/BukkitCommandPreprocessor.java
```diff
@@ -76,5 +79,22 @@ public void accept(final @NonNull CommandPreprocessingContext<C> context) {
                 BukkitCommandContextKeys.CLOUD_BUKKIT_CAPABILITIES,
                 this.commandManager.queryCapabilities()
         );
+
+        // Store if PaperCommandManager's preprocessor didn't already
+        if (!context.commandContext().contains(BukkitCommandContextKeys.SENDER_SCHEDULER_EXECUTOR)) {
```
Citymonstret, line 84: This should probably use [computeIfAbsent](https://github.com/Incendo/cloud/blob/2.0.0-dev/cloud-core/src/main/java/cloud/commandframework/keys/MutableCloudKeyContainer.java#L154)

PR #600, opened 2023-12-27
Make CommandSuggestionProcessors operate on a stream

jpenilla:
this allows filter/sort/limit without collecting multiple times

### DIFF
--- a/cloud-core/src/main/java/cloud/commandframework/execution/FilteringCommandSuggestionProcessor.java
+++ b/cloud-core/src/main/java/cloud/commandframework/execution/FilteringCommandSuggestionProcessor.java
@@ -29,8 +29,10 @@ import cloud.commandframework.internal.CommandInputTokenizer;
 import java.util.Iterator;
 import java.util.List;
 import java.util.Locale;
+import java.util.Objects;
 import java.util.function.BiFunction;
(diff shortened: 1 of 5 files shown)


cloud-core/src/main/java/cloud/commandframework/execution/FilteringCommandSuggestionProcessor.java
```diff
@@ -67,27 +69,29 @@ public FilteringCommandSuggestionProcessor(final @NonNull Filter<C> filter) {
...
-            return null;
-        }
-        return suggestion.withSuggestion(filtered);
+        return suggestions.map(suggestion -> {
+            final String input;
+            if (context.commandInput().isEmpty(true /* ignoreWhitespace */)) {
+                input = "";
+            } else {
+                input = context.commandInput().skipWhitespace().remainingInput();
+            }
```
Citymonstret, line 82:
```suggestion
        final String input;
        if (context.commandInput().isEmpty(true /* ignoreWhitespace */)) {
            input = "";
        } else {
            input = context.commandInput().skipWhitespace().remainingInput();
        }
        return suggestions.map(suggestion -> {
```

Citymonstret (approved):

PR #604, opened 2023-12-28
feat(brigadier): add missing generic to BrigadierManagerHolder, switc…

jpenilla:
…h #brigadierManager nullability for a boolean method and exception

Checking for null is fine when working with generic code, but when working directly with a PaperCommandManager for example, it becomes annoying as we must have already checked capabilities and registered brigadier before calling #brigadierManager.

### DIFF
--- a/cloud-minecraft/cloud-velocity/src/main/java/cloud/commandframework/velocity/VelocityCommandManager.java
+++ b/cloud-minecraft/cloud-velocity/src/main/java/cloud/commandframework/velocity/VelocityCommandManager.java
@@ -172,8 +172,9 @@ public class VelocityCommandManager<C> extends CommandManager<C> implements Brig
      * @return {@code true}
      * @since 2.0.0
      */
+    @API(status = API.Status.STABLE, since = "2.0.0")
     @Override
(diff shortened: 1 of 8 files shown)


cloud-minecraft/cloud-velocity/src/main/java/cloud/commandframework/velocity/VelocityCommandManager.java
```diff
@@ -165,13 +166,26 @@ public final boolean hasPermission(final @NonNull C sender, final @NonNull Strin
...
-     * In the case of the {@link VelocityCommandManager}, Brigadier is always used for command registration,
-     * and therefore this method will never return {@code null}.
      *
+     * <p>This will always return true for {@link VelocityCommandManager}s.</p>
+     *
+     * @return {@code true}
+     * @since 2.0.0
+     */
+    @Override
+    public boolean hasBrigadierManager() {
```
Citymonstret, line 176: should this be final? It makes no sense to override it. Same deal wit fabric. Do we want `@API` annotations these?

Citymonstret (approved):
lgtm 🍍

PR #613, opened 2024-01-01
Add SenderMapper type to wrap map & reverse map functions

jpenilla:


cloud-core/src/main/java/cloud/commandframework/SenderMapperHolder.java
```diff
@@ -0,0 +1,44 @@
...
+package cloud.commandframework;
+
+import org.apiguardian.api.API;
+import org.checkerframework.checker.nullness.qual.NonNull;
+
+/**
+ * A holder of a {@link SenderMapper}.
+ *
+ * @param <B> base command sender type
+ * @param <M> mapped command sender type
```
Citymonstret, line 33:
```suggestion
 * @param <M> mapped command sender type
 * @since 2.0.0
```

Citymonstret (commented):
Looks good. It makes things a lot cleaner :) Found a single missing `@since` :cow:

Citymonstret (approved):

PR #629, opened 2024-01-11
Respect sender types & permissions in more places

jpenilla:
- Propagate required sender types down the command tree and use that information for Brigadier requirement checks
- Check permissions and sender types in the StandardCommandSyntaxFormatter
- Check sender types before suggesting

todo:
- use for help? (are there any other spots I'm missing?)
- more tests?

### DIFF
--- a/cloud-core/src/main/java/cloud/commandframework/arguments/StandardCommandSyntaxFormatter.java
+++ b/cloud-core/src/main/java/cloud/commandframework/arguments/StandardCommandSyntaxFormatter.java
@@ -24,12 +24,17 @@
 package cloud.commandframework.arguments;
 
 import cloud.commandframework.CommandComponent;
+import cloud.commandframework.CommandManager;
 import cloud.commandframework.arguments.aggregate.AggregateCommandParser;
 import cloud.commandframework.arguments.flags.CommandFlag;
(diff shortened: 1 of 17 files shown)


cloud-core/src/main/java/cloud/commandframework/arguments/StandardCommandSyntaxFormatter.java
```diff
@@ -40,20 +45,64 @@
...
 public class StandardCommandSyntaxFormatter<C> implements CommandSyntaxFormatter<C> {
 
+    private final CommandManager<C> manager;
+
+    /**
+     * Creates a new {@link StandardCommandSyntaxFormatter}.
+     *
+     * @param manager command manager
+     */
+    public StandardCommandSyntaxFormatter(final CommandManager<C> manager) {
```
Citymonstret, line 64:
```suggestion
    public StandardCommandSyntaxFormatter(final @NonNull CommandManager<C> manager) {
```

Citymonstret (commented):
Good idea to propagate the sender type alongside the permission. This solves the issue in cloud-brigadier very nicely :+1:

Citymonstret (approved):

PR #634, opened 2024-01-13
feat(core): make `CommandComponent` immutable

Citymonstret:
The "owning command" has been moved into the command node, as the node is always present when we access the command anyway. This allows us to only have final fields in the command component.

This also means that we no longer need to copy the components before storing them in the command tree, and we can instead re-use the components.


jpenilla (approved):

cloud-core/src/main/java/cloud/commandframework/Command.java
```diff
@@ -2252,12 +2248,15 @@ private Builder(
...
-                final CommandComponent<C> componentCopy = component.copy();
-                builder = builder.argument(componentCopy);
+                builder = builder.argument(component);
             }
             if (this.permission.permissionString().isEmpty()) {
                 builder = builder.permission(command.commandPermission());
             }
+            final Class<? extends C> senderType = command.senderType().orElse(null);
+            if (senderType != null) {
+                builder = builder.senderType(senderType);
```
jpenilla, line 2258: shouldn't the resulting builder copy the generic type of `command`?

jpenilla: should probably make this change in a separate PR anyways

jpenilla (approved):

PR #635, opened 2024-01-13
chore: remove "hidden" functionality

Citymonstret:
This isn't used anywhere. The other alternative is that we do make use of this, but then we need to decide _where_ the command is hidden. In help menus? In syntax strings? Suggestions?


jpenilla (approved):

zml2008:
I have something like this in `adventure-platform-fabric` to hide the `/callback` command from client completion: https://github.com/KyoriPowered/adventure-platform-fabric/blob/mc/1.20/src/main/java/net/kyori/adventure/platform/fabric/impl/HiddenRequirement.java#L38, applied with [this mixin](https://github.com/KyoriPowered/adventure-platform-fabric/blob/3a6f2f7803ba381a1541c88a0b5c31681deaf35c/src/mixin/java/net/kyori/adventure/platform/fabric/impl/mixin/minecraft/commands/CommandsMixin.java#L95)

Would it be useful to preserve this feature and give it that sort of functionality?

jpenilla:
I think it would be better to reimplement that from scratch then to keep around this unused bloat until a point when someone implements it

PR #669, opened 2024-01-27
Allow parsing multiple instances

masmc05:
This allows parsing multiple instances at once with better support of complex relations between instances (like if instance A uses suggestions from B and at the same time B uses suggestions from A)

### DIFF
--- a/cloud-annotations/src/main/java/org/incendo/cloud/annotations/AnnotationParser.java
+++ b/cloud-annotations/src/main/java/org/incendo/cloud/annotations/AnnotationParser.java
@@ -721,19 +721,22 @@ public final class AnnotationParser<C> {
      * @return Collection of parsed commands
      */
     @SuppressWarnings({"unchecked", "rawtypes"})
-    public <T> @NonNull Collection<org.incendo.cloud.@NonNull Command<C>> parse(final @NonNull T... instances) {
-        for (T instance: instances) {
+    public <T> @NonNull Collection<org.incendo.cloud.@NonNull Command<C>> parse(final @NonNull T @NonNull... instances) {
+        for (final T instance : instances) {
+            this.parseDefaultValues(instance);
+        }
+        for (final T instance : instances) {
             this.parseSuggestions(instance);
         }
-        for (T instance: instances) {
+        for (final T instance : instances) {
             this.parseParsers(instance);
         }
-        for (T instance: instances) {
+        for (final T instance : instances) {
             this.parseExceptionHandlers(instance);
         }
 
         ArrayList<org.incendo.cloud.Command<C>> result = new ArrayList<>();
-        for (T instance: instances) {
+        for (final T instance : instances) {
             final Collection<CommandDescriptor> commandDescriptors = this.commandExtractor.extractCommands(instance);
             final Collection<org.incendo.cloud.Command<C>> commands = this.construct(instance, commandDescriptors);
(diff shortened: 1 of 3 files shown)


cloud-annotations/src/main/java/org/incendo/cloud/annotations/AnnotationParser.java
```diff
@@ -708,19 +708,43 @@ public void defaultValueRegistry(final @NonNull DefaultValueRegistry<C> defaultV
...
+     * Scan some classes instances of {@link Command} annotations and attempt to
+     * compile them into {@link org.incendo.cloud.Command} instances.
+     *
+     * @param instances Instances to scan
+     * @param <T>      Type of the instance
+     * @return Collection of parsed commands
+     */
+    @SuppressWarnings({"unchecked", "rawtypes"})
+    public <T> @NonNull Collection<org.incendo.cloud.@NonNull Command<C>> parse(final @NonNull T... instances) {
+        for (T instance: instances) {
```
jpenilla, line 725: The checkstyle not flagging this is concerning... `:` as part of enhanced for needs a single space on either side (`Type t : iterable`). `instance` should also be final.

masmc05: Fixed

cloud-annotations/src/main/java/org/incendo/cloud/annotations/AnnotationParser.java
```diff
@@ -708,19 +708,43 @@ public void defaultValueRegistry(final @NonNull DefaultValueRegistry<C> defaultV
...
+            this.parseSuggestions(instance);
+        }
+        for (final T instance : instances) {
+            this.parseParsers(instance);
+        }
+        for (final T instance : instances) {
+            this.parseExceptionHandlers(instance);
+        }
+
+        ArrayList<org.incendo.cloud.Command<C>> result = new ArrayList<>();
```
jpenilla, line 738:
```suggestion
        final List<org.incendo.cloud.Command<C>> result = new ArrayList<>();
```

jpenilla:
Should probably update parseContainers to use the new method

cloud-annotations/src/main/java/org/incendo/cloud/annotations/AnnotationParser.java
masmc05: Fixed

masmc05:
Updated the parseContainers, now if the priorities are the same, they will be parsed together in the new method.

```java
        @CommandContainer
        class Test1 {
            @Command("test1 <arg>")
            public void command(Object sender, @Argument(value = "arg", suggestions = "test2") String arg) {
            }

            @Suggestions("test1")
            public List<String> suggestions(Object sender, CommandInput arg) {
                return Collections.emptyList();
            }
        }
        @CommandContainer
        class Test2 {
            @Command("test2 <arg>")
            public void command(Object sender, @Argument(value = "arg", suggestions = "test1") String arg) {
            }

            @Suggestions("test2")
            public List<String> suggestions(Object sender, CommandInput arg) {
                return Collections.emptyList();
            }
        }
```
By default now this works, but if one of the priorities will be higher or lower (for example `@CommandContainer(priority = 2)`), it will be parsed as in older version

jpenilla (dismissed):
lgtm

jpenilla (approved):

PR #678, opened 2024-02-02
feat(core): add `futureHandler` to `Command.Builder`

### LINKED ISSUE #664: Add `handler` overloads for future-producing command execution handlers
To return a `CompletableFuture` in a command execution handler you currently have to do something along the lines of
```java
builder.handler((FutureCommandExecutionHandler<C>) context -> someFuture);
```

We should introduce new builder methods that accept `FutureCommandExecutionHandler>` specifically. I think the cleanest way would be to add `futureHandler`, but I am open to suggestions.

Citymonstret:
Fixes #664.

### DIFF
--- a/cloud-core/src/main/java/org/incendo/cloud/Command.java
+++ b/cloud-core/src/main/java/org/incendo/cloud/Command.java
@@ -2025,6 +2025,18 @@ public class Command<C> {
         }
 
         /**
+         * Specifies the command execution handler.
+         *
+         * @param commandExecutionHandler New execution handler
+         * @return new builder instance using the command execution handler
+         */
+        public @NonNull Builder<C> futureHandler(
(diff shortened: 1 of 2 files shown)


cloud-core/src/main/java/org/incendo/cloud/Command.java
```diff
@@ -2024,6 +2024,27 @@ private Builder(
...
+        /**
+         * Specifies the command execution handler.
+         *
+         * @param commandExecutionHandler New execution handler
+         * @return new builder instance using the command execution handler
+         */
+        public @NonNull Builder<C> futureHandler(
+                final CommandExecutionHandler.@NonNull FutureCommandExecutionHandler<C> commandExecutionHandler
+        ) {
+            return new Builder<>(
```
jpenilla, line 2036: why not just delegate to `handler`?

jpenilla (approved):

PR #719, opened 2024-04-07
feat(core): Replace compound arguments with specialized aggregate parser builders

### LINKED ISSUE #711: ArgumentPair should support TypeToken
Currently, argumentpair only supports Pair<Class, Class> which causes some issues where I have a type with a generic (wildcard) as one of the types. There are probably issues with methods having the same signature if you just added overloads that took pairs of typetokens.

jpenilla:
fixes #711 

also fixes the resulting type(token)s (literal `Pair<U extends Object, V extends Object>` -> actual type ie `Pair<String, Integer>`)


Machine-Maker:
Does it have to be setup such that you first have to add the parsers to the ParserRegistry? I pretty much never use that cause I don't use annotations. Being able to set the types+parser directly instead of having to go through 1 extra step would be nice.-

jpenilla:
I've added some factory methods that take ParserDescriptors, I'm not sure about adding even more methods to Command.Builder though.

Citymonstret (approved):

PR #729, opened 2024-05-04
feat: allow for customization of enum parsing

### LINKED ISSUE #680: Allow customizing the appearance of enums
It would be neat if there was an interface you could implement on enums to make the enum parser use alternative names for suggestions and parsing.

Something like
```java
public enum YourEnum implements ParseableEnum {
    FOO,
    BAR
    BAZ;

   @Override
   public @NonNull String parseableName() {
       return this.name().toLowercase( ) + "_suffix";
   }
}
```

Citymonstret:
Fixes #680.


Citymonstret:
Tried this out but it's clunkier than just making a custom parser.

PR #741, opened 2024-05-23
feat: Add convenience formatting caption methods.

NFRCG:
The following PR adds convenience methods to allow caption formatting from the CommandContext for a different recipient that is **not** the command sender.

The benefit of this is to allow usage of the same mechanisms for formatting/sending captions for the sender AND the recipient of a command.

If there is an alternative or preferred way to achieve the same goal, please let me know and I'll close this PR.


broccolai:
not sure about this one, you could just do this yourself by getting the CaptionFormatter / CaptionRegistry from the CommandManager. If you still think there's a good reason change this pr a little so that the old methods call this method and supply this.commandSender

PR #772, opened 2024-09-11
fix: `DurationParser` allowing invalid input

emilyy-dev:
`DurationParser` using that regex and `Matcher#find()` only finds the pattern in the given input string, while the pattern is right for individual segments denoting value-unit sequences, it does not prevent the input to have non-matching content.
This patch fixes that by manually parsing the input.

Additionally it conforms to the recommended way of parsing an argument: peek input, pop input on successful parsing.


jpenilla:
> Additionally it conforms to the recommended way of parsing an argument: peek input, pop input on successful parsing.

This isn't the case in 2.x, the tree will restore the cursor on a failed parse. You only need to do this if your parsing logic calls for it.

emilyy-dev:
> > Additionally it conforms to the recommended way of parsing an argument: peek input, pop input on successful parsing.
> 
> This isn't the case in 2.x, the tree will restore the cursor on a failed parse. You only need to do this if your parsing logic calls for it.

In that case the docs need updating 👍🏻 I'll revert that change here for now

jpenilla (dismissed):

Citymonstret:
> > > Additionally it conforms to the recommended way of parsing an argument: peek input, pop input on successful parsing.
> > 
> > 
> > This isn't the case in 2.x, the tree will restore the cursor on a failed parse. You only need to do this if your parsing logic calls for it.
> 
> In that case the docs need updating 👍🏻 I'll revert that change here for now

Yes, the docs should be updated. When we switched over to the `CommandInput` there wasn't exactly a plan to no longer require peeking. That was introduced later on to increase the comfort of using the new system. I haven't done a thorough sweep of the documentation to reflect this (although it is at the very least pointed out in the external docs now)

Citymonstret (dismissed):

Citymonstret (dismissed):

jpenilla (approved):

PR #783, opened 2024-11-18
fix(kotlin): propagate exceptions properly on failed parsing

masmc05:
The parser annotation javadoc guarantees that exceptions thrown by the method will be wrapped into a failure. Currently on enabling coroutines support this behavior disappears, exceptions being propagated to an internal exception. 

This pr fixed this by wrapping parser exceptions into failures.


Citymonstret:
This seems sane, but this definitely warrants a test to make sure there's no future regression

PR #791, opened 2025-05-20
Add Permission#append

powercasgamer:


Citymonstret:
This permission syntax is highly Minecraft-specific and does not make any real sense to add to cloud-core as it completely lacks significance in other contexts

